PR #1, opened 2021-02-20
Initial backend

isaaccormack:
Implemented the serverless functions as discussed in architecture diagram / chats on discord

### DIFF
--- a/firebase.json
+++ b/firebase.json
@@ -7,7 +7,8 @@
     "predeploy": [
       "npm --prefix \"$RESOURCE_DIR\" run lint",
       "npm --prefix \"$RESOURCE_DIR\" run build"
-    ]
+    ],
+    "source": "functions"
   },
   "hosting": {
     "public": "build",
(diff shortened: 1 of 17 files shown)


firebase.json
```diff
@@ -31,13 +32,19 @@
...
       "port": 5000
     },
     "ui": {
       "enabled": true
+    },
+    "database": {
+      "port": 9000
+    },
+    "pubsub": {
+      "port": 6000
```
aomi, line 47: This is to enable scheduled cloud function triggering?

functions/package.json
```diff
@@ -4,6 +4,7 @@
     "lint": "eslint --ext .js,.ts .",
     "lint:fix": "eslint '*/**/*.{js,ts}' --quiet --fix",
     "build": "tsc",
+    "test": "tsc && mocha --reporter spec",
```
aomi, line 7: considering CRA uses Jest I think we should keep it the same instead of bringing in Mocha? I'm not familiar with Mocha if it provides anything useful for testing cloud functions but.

firebase.json
isaaccormack: yeah, scheduled functions are pubsub.

functions/package.json
isaaccormack: I would love to use jest to test the cloud functions, however the firebase cloud function tutorial only illustrates how to test with mocha and there are some specific things in the tests done with mocha which would not trivially translate to Jest. I tried to use Jest instead initially but gave up after a lot of troubles. 

I think this is more of a nicety than necessity and isn't a useful place to spend time

PR #3, opened 2021-02-21
Setup front-end

keithradford:
All changes here are tentative, open to feedback on everything.

Created preliminary file structure for the front-end and got rid of some unnecessary files from `create-react-app`. The important files like `index.tsx` and `App.tsx` are still in `/src`.

Each main "container" has been put into it's own folder in the `/app` directory. I've split the page into 3 containers: content (main area with course content), header, and sidebar. I figured these are the containers which will be holding all of the pages content, so it should be sufficient that they have their own folders. Within each container folder, there are the following directories and files:

- `/__tests__`: for any tests
- `/api`: for all calls a specific container may make to the API (i.e. sidebar to load courses from Kuali)
- `/components`: for any components. A component is a "dumb" part of the page. For example, the cards on the sidebar that show each course matching search criteria, or the seats bar in the main content area.
- `/hooks`: for any custom hooks a component or the container may use.
- `/shared`: for files like `types.ts` or other information that may be shared across various files in the container
- `<Container>.tsx`: the main code for the container that ties all the components, API calls, etc together. This is where any API calls should be made and formatting of the container as a whole.

In terms of component libraries, we'd decided ant design and chakra could both be used in the project. Although honestly, I think we could easily just go with chakra. I'm also still considering setting up Storybook and adding it to this PR, still need to look into it.


src/index.tsx
```diff
@@ -1,17 +1,10 @@
...
     <App />
   </React.StrictMode>,
-  document.getElementById('root')
+  document.getElementById("root")
 );
-
-// If you want to start measuring performance in your app, pass a function
-// to log results (for example: reportWebVitals(console.log))
-// or send to an analytics endpoint. Learn more: https://bit.ly/CRA-vitals
-reportWebVitals();
```
aomi, line 17: We'll want to set this up so I'd prefer to leave it.

package.json
```diff
@@ -43,6 +43,14 @@
       "last 1 chrome version",
       "last 1 firefox version",
       "last 1 safari version"
-    ]
+    ],
+    "jest": {
+      "transformIgnorePatterns": [
+        "/node_modules/(?!antd|@ant-design|rc-.+?|@babel/runtime).+(js|jsx)$"
+      ]
+    }
```
aomi, line 51: wrong scope, this breaks `npm start`. move it outside.

aomi (approved):
One issue but otherwise :+1:

PR #4, opened 2021-02-22
Backend Initial Work

aomi:
This PR refactors #1 and implements the same functionality. 

:warning: probably not worth reading through the generated code!


functions/src/sections/Section.service.ts
```diff
@@ -0,0 +1,90 @@
...
+        await Promise.all(
+          terms.map(async (term) => {
+            const sections = await UVicCourseScraper.getCourseSections(
+              term,
+              subject,
+              code
+            );
+            // don't want to put empty mappings in the database
+            if (sections.length > 0) {
+              console.info(term, subject, code, sections.length);
```
isaaccormack, line 72: remove?

functions/src/app.ts
```diff
@@ -0,0 +1,36 @@
...
+app.use(
+  bodyParser.urlencoded({
+    extended: true,
+  })
+);
+app.use(bodyParser.json());
+
+// a slight "hack" to get things to play well in Cloud Functions
+// none of the below is required when running locally.
+if (process.env.NODE_ENV === 'production') {
```
isaaccormack, line 23: briefly, why do we need this?

functions/src/sections/Section.controller.ts
```diff
@@ -0,0 +1,43 @@
...
+import { Term } from '../constants';
+import { Section } from './Section.model';
+import { SectionsService } from './Section.service';
+
+@Route('sections')
+export class SectionsController extends Controller {
+  @Response(404, 'Section Not Found')
+  @Get('{term}')
+  public async sections(
+    @Path() term: Term,
```
isaaccormack, line 13: this throws a nasty validation error when the term isn't found, is there anyway to add some middleware to catch this error and give a nicer error back to the client like simply, invalid term. This is just a nice to have, feel free to say no.

Example of response for term=202000:

<!DOCTYPE html>
<html lang="en">

<head>
	<meta charset="utf-8">
	<title>Error</title>
</head>

<body>
	<pre>ValidateError: <br> &nbsp; &nbsp;at new ValidateError (/Users/i/Develop/clockwork/functions/node_modules/@tsoa/runtime/src/routeGeneration/templateHelpers.ts:791:5)<br> &nbsp; &nbsp;at getValidatedArgs (/Users/i/Develop/clockwork/functions/build/routes.ts:254:19)<br> &nbsp; &nbsp;at /Users/i/Develop/clockwork/functions/build/routes.ts:143:33<br> &nbsp; &nbsp;at Layer.handle [as handle_request] (/Users/i/Develop/clockwork/functions/node_modules/express/lib/router/layer.js:95:5)<br> &nbsp; &nbsp;at next (/Users/i/Develop/clockwork/functions/node_modules/express/lib/router/route.js:137:13)<br> &nbsp; &nbsp;at Route.dispatch (/Users/i/Develop/clockwork/functions/node_modules/express/lib/router/route.js:112:3)<br> &nbsp; &nbsp;at Layer.handle [as handle_request] (/Users/i/Develop/clockwork/functions/node_modules/express/lib/router/layer.js:95:5)<br> &nbsp; &nbsp;at /Users/i/Develop/clockwork/functions/node_modules/express/lib/router/index.js:281:22<br> &nbsp; &nbsp;at param (/Users/i/Develop/clockwork/functions/node_modules/express/lib/router/index.js:354:14)<br> &nbsp; &nbsp;at param (/Users/i/Develop/clockwork/functions/node_modules/express/lib/router/index.js:365:14)</pre>
</body>

</html>

isaaccormack (changes requested):
couple comments, other than that, this is awesome

functions/src/sections/Section.controller.ts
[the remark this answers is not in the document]
aomi: Yup

functions/src/courses/Course.controller.ts
```diff
@@ -0,0 +1,29 @@
+import { Controller, Get, Path, Query, Route } from 'tsoa';
+import { Course } from './Course.model';
+import { CoursesService } from './Course.service';
+
+@Route('courses')
+export class CoursesController extends Controller {
+  /**
+   * Retrieves all the courses available. If query params are passed in, they will be used to filter results.
```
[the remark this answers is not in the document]
aomi: My thinking is typically they exist to reduce what's being sent to the client. Given there's like 3400~ entries it would be nice to for the API to slim down the response if requested.

functions/src/courses/Course.controller.ts
```diff
@@ -0,0 +1,29 @@
...
+   * @param subject
+   * @param code
+   */
+  @Get()
+  public async getAllCourses(
+    @Query() subject?: string,
+    @Query() code?: string
+  ): Promise<Course[]> {
+    console.log(subject, code);
```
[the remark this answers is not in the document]
aomi: Oops :see_no_evil:

functions/src/sections/Section.controller.ts
aomi: There is a way but I haven't gotten around to it yet fully. It doesn't seem it should be a blocker though for this PR. In production that error isn't displayed to the client.

functions/src/app.ts
aomi: Since the CDN rewrite rule in `firebase.json` sends all requests that start with `/api` but the Express app doesn't have a route that starts with `/api`, we need to remove it or make another router with `app.use('/api', actualRouter)` and I opted for just rewriting instead.

aomi: But when it's running locally you don't need the `/api` prefix and the rewrite rule works differently.

functions/src/sections/Section.service.ts
```diff
@@ -0,0 +1,90 @@
...
+import { UVicCourseScraper } from '@isaaccormack/uvic-course-scraper';
+import { db } from '../db/firestore';
+import { DetailedClassInformation } from '@isaaccormack/uvic-course-scraper/dist/src/types';
+
+export class SectionsService {
+  public async getSections(
+    term: string,
+    subject: string,
+    code: string
+  ): Promise<Section[]> {
```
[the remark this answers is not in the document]
aomi: My idea is we shouldn't have a hard dependency on types from another library. More or less the attributes of `Section` will be very similar to `ClassScheduleListing` but I don't think it should be 1:1 as there might be customization we want to do on our end. It seems a bit safer and doesn't result in much more code and it's all still typed anyhow.

functions/src/courses/Course.controller.ts
```diff
@@ -0,0 +1,29 @@
...
+    // set the Cache-Control for 24h.
+    this.setHeader('Cache-Control', `public, max-age=${3600 * 24}`);
+    return new CoursesService().getAllCourses();
+  }
+
+  @Get('{pid}')
```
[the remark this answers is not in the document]
aomi: If/when we implement query params that allow the results to be altered then yes but if this endpoint fails I'd deem it not a 404 moreso a 500

functions/src/courses/Course.service.ts
```diff
@@ -0,0 +1,18 @@
+import { Course } from './Course.model';
+// TODO: replace once main repo is updated.
+import { UVicCourseScraper } from '@isaaccormack/uvic-course-scraper';
+
+export class CoursesService {
+  public async getAllCourses(): Promise<Course[]> {
+    const courses = await UVicCourseScraper.getAllCourses();
+    return courses.map((e) => ({
```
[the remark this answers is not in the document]
aomi: Rational is same as the `Section` thing. I just haven't added all the attributes so you're totally right. I'll add those and probs a few more.

functions/src/courses/Course.service.ts
```diff
@@ -0,0 +1,18 @@
...
+      id: e.pid,
+      title: e.title,
+    }));
+  }
+
+  public async getCourseDetailsByPid(pid: string): Promise<Course> {
+    const course = await UVicCourseScraper.getCourseDetailsByPid(pid);
+    return { id: course.pid, title: course.title };
```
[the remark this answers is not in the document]
aomi: Yeah that makes sense. I'll switch it :+1:

functions/src/sections/Section.controller.ts
isaaccormack: Ok

functions/src/app.ts
isaaccormack: Makes sense, thanks for explaining

functions/src/app.ts
```diff
@@ -0,0 +1,39 @@
...
+// to the Cloud Function but the Express app doesn't have a route that starts
+// with `/api`, we need to remove it or make another router with app.use('/api', actualRouter)`
+if (process.env.NODE_ENV === 'production') {
+  app.use((req, res, next) => {
+    if (req.url.indexOf(`/${API_PREFIX}/`) === 0) {
+      req.url = req.url.substring(API_PREFIX.length + 1);
+    }
+    next();
+  });
+  app.get('/openapi.json', async (req, res) => {
```
isaaccormack, line 33: why only use this in pro?

functions/src/sections/Section.service.ts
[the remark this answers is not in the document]
isaaccormack: This is clearly a simple tradeoff.

Seems easier just to use a hard dependency (don't need to rewrite and maintain a second type defn) but we lose flexibility. As we currently want to send everything (more or less) in ClassScheduleListing to the client I would argue we should just use it until the explicit use case arises where its better to customize our types.

functions/src/courses/Course.controller.ts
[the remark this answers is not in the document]
isaaccormack: Right. Seems like a 400 should be sent back if no course details found for the pid, but an empty list is good for now.

functions/src/courses/Course.controller.ts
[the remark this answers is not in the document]
isaaccormack: Can you describe a use case for this? I see what you're doing now, filter by either subject or code (or both) (ie. get all courses with subject="CSC").

functions/src/sections/Section.service.ts
```diff
@@ -0,0 +1,89 @@
...
+
+    return sections.map((e) => ({ crn: e.crn }));
+  }
+
+  public async getSectionSeats(
+    term: string,
+    subject: string,
+    code: string
+  ): Promise<DetailedClassInformation[]> {
+    console.log(term, subject, code);
```
isaaccormack, line 27: remove?

functions/src/sections/Section.service.ts
```diff
@@ -0,0 +1,89 @@
...
+    // get term, subject, code to crn mappings from db.
+    const doc = await db.sectionMappings
```
isaaccormack, line 29: might be nice here to bring subject.toUpper() before checking the DB so subject='csc' and subject='CSC' both work.

aomi: I'll add it at the controller level.

functions/src/app.ts
aomi: Moved it out!

functions/src/app.ts
aomi: I've added the blurb into the code as well.

functions/src/courses/Course.controller.ts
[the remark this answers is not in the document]
aomi: Fair point. We'll add it as we need when the use case comes up.

functions/src/sections/Section.service.ts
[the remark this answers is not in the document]
aomi: I strongly believe we shouldn't just expose the types from uvic-course-scraper directly as we don't want that hard dependency given the solution to avoiding that is so easy. For now, I just extended the type so we're "technically" dependency on it but there's a level of separation now.

functions/src/courses/Course.service.ts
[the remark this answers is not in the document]
aomi: I added those and other attributes.

functions/src/sections/Section.service.ts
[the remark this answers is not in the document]
isaaccormack: Ok, seems like a good compromise for now

functions/package.json
```diff
@@ -8,28 +8,42 @@
...
     "deploy": "firebase deploy --only functions",
-    "logs": "firebase functions:log"
+    "logs": "firebase functions:log",
+    "dev": "concurrently \"nodemon\" \"nodemon -x tsoa spec-and-routes\"",
+    "build:routes": "tsoa spec-and-routes && tsc",
+    "start:server": "node build/src/server.js"
   },
   "engines": {
-    "node": "12"
+    "node": "14"
```
isaaccormack, line 17: do we have to move up from 12? would be nice to be consistent with course-scraper so we don't need to change node versions between developing?

functions/package.json
```diff
@@ -8,28 +8,42 @@
...
   },
-  "main": "lib/index.js",
+  "main": "lib/src/index.js",
   "dependencies": {
+    "@isaaccormack/uvic-course-scraper": "^0.1.5",
```
isaaccormack, line 21: reminder: update this to vikelabs

functions/src/courses/Course.model.ts
```diff
@@ -0,0 +1,32 @@
+import {
+  KualiCourseCatalog,
+  KualiCourseItem,
+} from '@vikelabs/uvic-course-scraper/dist/types';
+
+export interface Course
+  extends Pick<KualiCourseCatalog, 'pid' | 'title' | 'dateStart'> {
```
isaaccormack, line 7: oooh very cool!

functions/src/index.ts
```diff
@@ -1,15 +1,59 @@
...
+
+import * as bodyParser from 'body-parser';
+import { RegisterRoutes } from '../build/routes';
+
+import * as openapi from '../build/swagger.json';
+
+// THIS FILE SHOULDN'T RUN WITHOUT FIREBASE EMULATOR!
+// nor is it meant to be. use server.ts!
+
+export const app = express();
```
isaaccormack, line 17: shouldn't this just import app from app.ts and use that? (ie. avoid this code duplication)

functions/src/sections/Section.controller.ts
```diff
@@ -0,0 +1,47 @@
+// TODO: replace package import path
```
isaaccormack, line 1: remove

functions/src/index.ts
aomi: yeahhhhhhhhhhhhhhhhhhhhhhhhhh so i like broke the app last night and I realized the middle-ware for rewriting URLs has to be loaded in before the routes and what no so I decided for now, till there's better tooling to differentiate run-time environments.

functions/src/sections/Section.service.ts
```diff
@@ -0,0 +1,87 @@
+import { Seat, Section } from './Section.model';
+// TODO: replace once main repo is updated.
```
isaaccormack, line 2: remove

functions/src/sections/Section.controller.ts
```diff
@@ -0,0 +1,47 @@
...
+    return sections;
+  }
+
+  @Response(404, 'Section Seats Not Found')
+  @Get('{term}/seats')
+  public async seats(
+    @Path() term: Term,
+    @Query() subject: string,
+    @Query() code: string
+  ): Promise<DetailedClassInformation[]> {
```
isaaccormack, line 31: should be Promise<Seat[]>

functions/src/sections/Section.model.ts
```diff
@@ -0,0 +1,16 @@
+import {
+  ClassScheduleListing,
+  DetailedClassInformation,
+} from '@vikelabs/uvic-course-scraper/dist/types';
+
+export interface Section extends ClassScheduleListing {
+  crn: string;
```
isaaccormack, line 7: ClassScheduleListing already has crn?

functions/src/sections/Section.service.ts
```diff
@@ -0,0 +1,87 @@
...
+  ): Promise<Seat[]> {
+    // get term, subject, code to crn mappings from db.
+    const doc = await db.sectionMappings
+      .doc(SectionsService.constructSectionKey(term, subject, code))
+      .get();
+
+    const data = doc.data();
+
+    if (!data) {
+      throw new InvalidSubjectCodeError('Section Seats Not Found');
```
isaaccormack, line 36: First, do you mean invalid subject and code pair?

This error can be thrown when the subject-code (course) is actually valid but the course just isn't in the db that term because there are no offerings.

aomi: We don't have a way to differentiate I guess by assuming a hit/miss in the database. We would have to make empty entries for valid queries but I don't know if that's a good idea or not...

aomi: In terms of the `error`, it was me just testing and leaving something to work off of.

functions/src/courses/Course.service.ts
```diff
@@ -0,0 +1,27 @@
+import { Course, CourseDetails } from './Course.model';
+import { UVicCourseScraper } from '@vikelabs/uvic-course-scraper/dist/index';
+import { subjectCodeExtractor } from '../shared/subjectCodeExtractor';
+
+export class CoursesService {
+  public async getAllCourses(): Promise<Course[]> {
+    const courses = await UVicCourseScraper.getAllCourses();
+    return courses.map((course) => ({
+      ...subjectCodeExtractor(course),
```
isaaccormack, line 9: very clean

isaaccormack (approved):

PR #19, opened 2021-02-25
Cache section mappings on a per-requested basis

aomi:


functions/src/sections/Section.service.ts
```diff
@@ -26,24 +26,57 @@ export class SectionsService {
...
+      const sections = await UVicCourseScraper.getCourseSections(
+        term,
+        subject,
+        code
+      );
+      if (sections.length > 0) {
+        const crns = sections.map(({ crn }) => crn);
+        await db.courseMappings
+          .doc(SectionsService.constructSectionKey(term, subject, code))
+          .set({ crns });
```
isaaccormack, line 76: shouldn't retrievedAt be written to database?

functions/src/sections/Section.service.ts
```diff
@@ -26,24 +26,57 @@ export class SectionsService {
...
+        return { crns, retrieveAt: new Date(Date.now()) };
+      }
+    } catch (e) {}
   }
 
   public async updateSectionMappings(): Promise<void> {
```
isaaccormack, line 82: should we just delete this then?

functions/src/sections/Section.service.ts
isaaccormack: should we have any error handling in case this throws? at least log the error? I'm open for discussion

isaaccormack (changes requested):
Good work, I think this is objectively better than using updateSectionMappings seeing how simple the implementation is, just a couple things...

functions/src/sections/Section.service.ts
aomi: Fixed the first comment, nice catch and thanks. Second, I did add some logging as it's better than nothing now. I haven't had time to looking into our monitoring ability of app yet but that's a whole other thing and out of scope for this task.

functions/src/sections/Section.service.ts
aomi: I don't think so. I can see this being useful in some situations or tasks during development.

isaaccormack: we could just delete it here and then copy it in from the git history if we ever need it so we don't have dead code...

functions/src/sections/Section.service.ts
isaaccormack: fair point

functions/src/sections/Section.service.ts
aomi: While we're still experimenting with caching so I'd prefer to leave it in.

functions/src/sections/Section.service.ts
aomi: I think it's a matter of adding middleware in Express but like I said, haven't found a good way.

isaaccormack (approved):
LGTM

PR #20, opened 2021-02-25
Setup Restful React

aomi:
Sets up [resful-react](https://github.com/contiamo/restful-react/) and provides an example of using it. 

restful-react generates typed-fetchers as react hooks using the OpenAPI 3 spec from the cloud functions. 

You might need something like https://chrome.google.com/webstore/detail/allow-cors-access-control/lhobafahddgcelffkeicbaginigeejlf to get around some CORS things.


keithradford (approved):
LGTM

PR #21, opened 2021-02-26
Add Header

keithradford:
Screen Shot 2021-02-25 at 6 23 46 PM

## Details

Added the header container to the main app. This includes: the "logo", a search bar, buttons to select the term, and a user icon for when that functionality is added.

The search bar uses the ant-design [component](https://ant.design/components/input/). It currently has a static width, but I have plans to make the page (or at least this component) more responsive once this is merged. 
Paired with the search bar is three buttons to select which term you are searching in. These buttons are dynamic in the sense that they display different terms in different locations depending on when the user accesses the site as well as update the year automatically. The selected term is a state from the highest level of the app so that data can be used outside the header.

## To do after merge
All of these could be done in this PR, but I want to get this basic functionality merged to set precedent for others doing tickets.
- [ ] Write tests
- [x] Make more responsive
- [ ] Make logo
- [ ] change colours?


isaaccormack:
I guess I would argue you don't need to merge the basic functionality to master to set a precedent for others as they can just look at this branch?

src/app/header/Header.tsx
```diff
@@ -1,24 +1,37 @@
...
-    <div>
-      {content}
-    </div>
-  )
-};
\ No newline at end of file
+    <Box bg="tomato" h={66} px="10" py="4">
+      <Flex minW="900px">
+        {/* TODO: turn this into a logo */}
+        <Box w={225} textAlign="center">
```
aomi, line 22: I would avoid setting an absolute value for width here.

src/app/header/Header.tsx
```diff
@@ -1,24 +1,37 @@
...
+          <Text fontSize="x-large" color="white" fontWeight="bold">
+            clockwork
+          </Text>
+        </Box>
+        <Spacer />
```
aomi, line 27: As said in the Discord, I don't think these spacer elements would be good to use in the header.

src/app/header/Header.tsx
```diff
@@ -1,24 +1,37 @@
...
+  setTerm: React.Dispatch<React.SetStateAction<string>>;
 }
 
 /**
  * Primary UI component for content
  */
-export const Header: React.FC<HeaderProps> = ({
-  content
-}) => {
+export const Header: React.FC<HeaderProps> = ({ setTerm }) => {
```
aomi, line 17: @zthh626 and I were dicussing and after some Googling around we came to the conclusion that the normal `function ({}: Props): JSX.Element` is the way to go.

src/app/header/Header.tsx
```diff
@@ -1,24 +1,37 @@
...
   return (
-    <div>
-      {content}
-    </div>
-  )
-};
\ No newline at end of file
+    <Box bg="tomato" h={66} px="10" py="4">
```
aomi, line 19: Same idea with the absolute height if we can avoid it.

src/app/header/Header.tsx
keithradford: yeah I prefer that too

src/app/header/shared/utils.ts
```diff
@@ -0,0 +1,49 @@
...
+    terms.push(`Summer ${nextYear}`);
+  }
+
+  return terms;
+}
+
+export function getTerms(date: dayjs.Dayjs = dayjs()): string[] {
+  const year = date.year().toString();
+  const prevYear = (date.year() - 1).toString();
+  const nextYear = (date.year() + 1).toString();
```
aomi, line 30: You don't need to output as strings since you use them as strings anyways.

src/app/header/shared/utils.ts
```diff
@@ -0,0 +1,49 @@
...
+  const currMonth = date.month();
+  const terms: string[] = [];
+
+  if (0 <= currMonth && currMonth < 4) {
+    terms.push(`${prevYear}09`);
+    terms.push(`${year}01`);
+    terms.push(`${year}05`);
```
aomi, line 37: you can return an array here directly with the items you're pushing into `terms` right now.

src/app/header/shared/utils.ts
```diff
@@ -0,0 +1,49 @@
+import dayjs from 'dayjs';
+
+export function getButtonTerms(date: dayjs.Dayjs = dayjs()): string[] {
```
aomi, line 3: The only difference between these two functions is the final output. I would maybe suggest writing a format function and try and re-use a lot of the logic (checking the ranges for month etc)

src/app/header/components/TermButtons.tsx
```diff
@@ -0,0 +1,67 @@
...
+export function TermButtons() {
+  const [term, setTerm] = useTerm();
+  const status = useMemo(
+    () => ({
+      first: true,
+      second: false,
+      third: false,
+    }),
+    []
+  );
```
aomi, line 16: I would prefer we use an array or object for this...

src/app/header/components/TermButtons.tsx
```diff
@@ -0,0 +1,67 @@
...
+  const terms = getTerms();
+
+  const onClick = useCallback(
+    (event: React.MouseEvent<HTMLElement, MouseEvent>, name: string) => {
+      event.preventDefault();
+      setTerm(name);
+      if (name === terms[0] && !status.first) {
+        status.first = true;
+        status.second = false;
+        status.third = false;
```
aomi, line 28: There _has_ to be a way to set every key value to a certain value and change the one to true...

src/app/header/components/TermButtons.tsx
```diff
@@ -0,0 +1,67 @@
...
+
+  return (
+    <ButtonGroup spacing="0" isAttached>
+      <Button
+        name={terms[0]}
+        isActive={status.first}
+        onClick={(e) => onClick(e, terms[0])}
+        size="sm"
+        borderRadius="2px"
+      >
```
aomi, line 50: I feel we should be able to iterate the buttons being rendered if we switch to an array of terms hopefully.

src/app/header/components/TermButtons.tsx
```diff
@@ -0,0 +1,67 @@
+import { Button, ButtonGroup } from '@chakra-ui/react';
+import React, { useCallback, useContext, useMemo } from 'react';
+
+import { useTerm } from '../../context/TermContext';
+import { getButtonTerms, getTerms } from '../shared/utils';
+
+export function TermButtons() {
+  const [term, setTerm] = useTerm();
```
aomi, line 8: I'm not sure if we want to store `term` here. I would think maybe storing and setting it in a context would be good (since a lot of stuff uses `term`)

src/app/header/Header.tsx
```diff
@@ -1,17 +1,31 @@
...
+            clockwork
+          </Text>
+        </Center>
+        <Spacer />
+        <HStack spacing="10px" w="60%">
+          <SearchBar />
+          <TermButtons />
+        </HStack>
+        <Spacer />
+        <UserButton />
```
aomi, line 27: I think we can hide this for now given we don't have any immediate plans for user profile stuff

src/app/context/TermContext.tsx
```diff
@@ -0,0 +1,22 @@
...
+
+export const useTerm = (): [string, Dispatch<SetStateAction<string>>] => {
+  return [useContext(TermContext), useContext(UpdateTermContext)];
+};
+
+type Props = {
+  children: JSX.Element;
+};
+
+export function TermProvider({ children }: Props) {
```
aomi, line 14: `React.PropsWithChild` is a thing.

src/app/header/components/TermButtons.tsx
aomi: Wait ignore me I misread.

src/app/shared/utils/terms.ts
```diff
@@ -0,0 +1,59 @@
...
+ * Fetches the readable text version of a term.
+ *
+ * @param {string} term i.e. '202101'
+ * @return {string} 'Spring 2021'
+ */
+export function getReadableTerm(term: string): string {
+  const yearRegex = /\d{4}/g;
+  const monthRegex = /(\d{4})(\d{2})/;
+  const year = yearRegex.exec(term)![0];
+  let month = monthRegex.exec(term)![2];
```
aomi, line 13: I think you can just use the length to get your required values and just note in the documentation you don't support anything otherwise. It's not like the current implementation fails gracefully. I was thinking just using `slice` to grab the values of year/month?

src/app/shared/utils/terms.ts
```diff
@@ -0,0 +1,59 @@
...
+/**
+ * Fetches the current terms of a school year given a date.
+ *
+ * @param {dayjs.Dayjs} date
+ * @return {string[]} array of terms ['202009', '202101', '202105']
+ */
+export function getCurrentTerms(date: dayjs.Dayjs = dayjs()): string[] {
+  const year = date.year().toString();
+  const prevYear = (date.year() - 1).toString();
+  const nextYear = (date.year() + 1).toString();
```
aomi, line 31: these don't need to have `toString()`.

src/app/header/components/TermButtons.tsx
```diff
@@ -0,0 +1,42 @@
...
+import React, { useCallback, useState } from 'react';
+
+import { useTerm } from '../../context/TermContext';
+import { getCurrentTerms, getReadableTerm } from '../../shared/utils/terms';
+
+export function TermButtons(): JSX.Element {
+  const [, setTerm] = useTerm();
+  const [status, setStatus] = useState([true, false, false]);
+
+  const terms = getCurrentTerms();
```
aomi, line 11: I feel all three can be combined into a utility hook called `useTerm`. It could maybe have the follow return attributes like `currentTerm, prevTerm, nextTerm, setTerm` etc. The `currentTerm` for example could be structured like `selected: true, formatted: '202101'` etc?

src/app/header/components/TermButtons.tsx
```diff
@@ -0,0 +1,42 @@
...
+      const status = [false, false, false];
+      status[i] = true;
+      setStatus(status);
+    },
+    [setTerm, status, terms]
+  );
+
+  return (
+    <ButtonGroup spacing="0" isAttached>
+      {terms.map((term, i) => {
```
aomi, line 26: sooooooooooooooooo much better :heart:

src/app/context/TermContext.tsx
```diff
@@ -0,0 +1,20 @@
...
+const UpdateTermContext = createContext<Dispatch<SetStateAction<string>>>(() => undefined);
+
+export const useTerm = (): [string, Dispatch<SetStateAction<string>>] => {
+  return [useContext(TermContext), useContext(UpdateTermContext)];
+};
+
+type Props = {};
+
+export function TermProvider({ children }: React.PropsWithChildren<Props>): JSX.Element {
+  const [term, setTerm] = useState('');
```
aomi, line 13: should we just allow term to be undefined?

aomi (commented):
love the refactor of the buttons :100:

src/app/shared/utils/terms.ts
keithradford: fair enough 👍

keithradford:
Screen Shot 2021-03-06 at 12 24 07 PM
updated screenshot as it's changed since opening PR

src/app/header/components/TermButtons.tsx
```diff
@@ -0,0 +1,42 @@
...
+        return (
+          <Button
+            key={i}
+            name={term}
+            isActive={status[i]}
+            onClick={(e) => onClick(e, term, i)}
```
aomi, line 32: It seems like a bad idea to use `useCallback` but also have it be making an anonymous function each time. It would be nice to get rid of either passing in `term` or `i`. Using `date-*` atrributes and accessing them through `e.currentTarget.getAttribute('date-*'')` might be a way around that. then you can just pass `onClick` directly (and avoid making a new function each time (and get max value for that `useCallback` imo)

src/app/shared/hooks/useTerm.ts
```diff
@@ -0,0 +1,13 @@
+import { createContext, Dispatch, SetStateAction, useContext } from 'react';
+
+import { getCurrentTerm } from '../utils/terms';
+
+const TermContext = createContext(getCurrentTerm());
+const UpdateTermContext = createContext<Dispatch<SetStateAction<string>>>(() => undefined);
```
aomi, line 6: I don't 100% understand the reason we need a "update context".

keithradford: from the context tutorial I watched and afaik this is what you do when you have a variable context, since we wanna change the term context that's why i added it

src/app/header/components/TermButtons.tsx
```diff
@@ -0,0 +1,38 @@
...
+    const name = event.currentTarget.getAttribute('name');
+    let idx = -1;
+    if (name) {
+      setTerm(name);
+      idx = terms.indexOf(name);
+    }
+    const status = [false, false, false];
+    status[idx] = true;
+    setStatus(status);
+  };
```
aomi, line 25: This should be in the if statement since if it isn't there's a chance it'll index negative.

aomi (approved):

PR #23, opened 2021-02-27
Feat/course card

chsunwoo1002:


zthh626:
@chsunwoo1002 Can we add some documentation to the props? For example:
`
export interface CardProps {
   title: string;
   subject: string;
  /**
   * The course code to be displayed
   *  Ex/ CSC 226 -> 226
   */
   code: string;
   selected?: boolean;
 }
`

src/app/sidebar/components/Card.tsx
```diff
@@ -0,0 +1,22 @@
...
+export interface CardProps {
+  title: string;
+  subject: string;
+  code: string;
+  selected?: boolean;
+}
+
+export function Card(props: CardProps): JSX.Element {  
+  return (
+    <Box bgColor={props.selected ? 'orange.200' : 'white'} color="black" m="5" p="4">
```
zthh626, line 13: Something really picky, but we should stick to either using ' or "

src/app/sidebar/components/Card.tsx
```diff
@@ -0,0 +1,22 @@
...
+import React from "react";
+
+export interface CardProps {
+  title: string;
+  subject: string;
+  code: string;
+  selected?: boolean;
+}
+
+export function Card(props: CardProps): JSX.Element {
```
zthh626, line 11: @aomi I usually build components like this, so that I don't need to call props.SOMETHING every time. Do you have a preference?
```
export const Card: React.FC<CardProps> = ({
  title,
  subject,
  code,
  selected
}) => {
```

aomi: I would personally go with 
```tsx
export const Card = ({title}: CardProps): JSX.Element => {};
```
if we wanted to use arrow functions but.

aomi: I mainly told Sunwoo to write it this way as I found it the simplest to explain as functions are pretty straightforward.

src/app/sidebar/components/Card.tsx
aomi: inb4 prettier. we'll fix it in a single go i think

src/app/sidebar/components/Card.tsx
zthh626: It seems like it's best practice to leave it the way it is: https://github.com/facebook/create-react-app/pull/8177

zthh626 (approved):

PR #25, opened 2021-02-28
Feat/course card drop down

zthh626:
Branched from @chsunwoo1002's card branch - Moved all colors to shared.styles.tsx (even though they are simple colors its I think its preferred to have them in one spot)

Create CardDropDown Component, takes these props:

- Subject
- Title
- Courses (List of Cards)

https://user-images.githubusercontent.com/31324580/109410533-452f3b80-7950-11eb-98ca-e4eebcfe2613.mov


zthh626:
What it looks like with box shadows.

Screen Shot 2021-02-27 at 11 30 12 PM

package.json
```diff
@@ -4,6 +4,7 @@
   "private": true,
   "homepage": ".",
   "dependencies": {
+    "@chakra-ui/icons": "^1.0.5",
```
aomi, line 7: Where is the `package-lock.json`?

src/app/sidebar/shared/styles.ts
```diff
@@ -0,0 +1,6 @@
+export const Colors = {
+    black: "black",
+    grey: "grey",
+    orange200: "#FBD38D",
+    white: "white"
+}
```
aomi, line 6: I get the idea and agree we shouldn't be using random hex and color constants however those values are actually fully typed and pulled from a global theme. Maybe we should define a set of colours but since we haven't using the built-in ones seems fine for now. 

This has documentation of how the global theme (where something like what you defined), might make more sense. 
https://chakra-ui.com/docs/features/global-styles

src/app/sidebar/components/CardDropDown.tsx
```diff
@@ -0,0 +1,64 @@
...
+const StyledBox = styled(Box)`
+    background-color: ${Colors.white};
+    color: ${Colors.black};
+    padding: 1em;
+    margin-top: 1em;
+`
+
+const StyledIconButton = styled(IconButton)`
+    padding: 0.5em;
+`
```
aomi, line 64: You can probably just pass do `<IconButton p="0.5em"/>` it's only one thing.

src/app/sidebar/components/CardDropDown.tsx
```diff
@@ -0,0 +1,64 @@
...
+        </Flex>
+    );
+}
+
+const StyledBox = styled(Box)`
+    background-color: ${Colors.white};
+    color: ${Colors.black};
+    padding: 1em;
+    margin-top: 1em;
+`
```
aomi, line 60: Same idea here. I don't think we should be pulling out `styled` unless we're implementing a complex design but that's up for debate.

src/app/sidebar/components/CardDropDown.tsx
```diff
@@ -0,0 +1,64 @@
...
+    */
+    title: string;
+    /*
+    * The list of courses for this subject
+    */
+    courseCards: JSX.Element[];
+}
+
+export function CardDropDown(props: CardProps): JSX.Element {
+    const [isDisplayed, setDisplayed] = useState(false);
```
aomi, line 25: I'm not sure if we want this component to hold the state. I think maybe that should be reflected in the main siderbar component. I think exposing a `onClick/onChange` etc. optional prop instead which would pass in what the user clicked (`pid`) would be better.

src/app/sidebar/components/CardDropDown.tsx
```diff
@@ -0,0 +1,64 @@
...
+    const handleClick = () => {
+        setDisplayed(!isDisplayed)
+    }
+
+    return (
+        <Flex direction="column">
+            <StyledBox boxShadow="lg">
+                <Flex direction="row" alignItems="center" justifyContent="space-between">
+                    <Heading size="sm" color={Colors.black} fontWeight={"normal"}>
+                        {props.subject} - {props.title}
```
aomi, line 36: Not in the Figma but I feel the title and subject should have a bit more contrast between them even if it's subtle.

src/app/sidebar/components/CardDropDown.tsx
zthh626: We talked about this offline - keeping it as state

src/app/sidebar/components/CardDropDown.tsx
zthh626: Screen Shot 2021-02-28 at 5 09 15 PM

aomi (approved):

isaaccormack:
just wanted to say, screen recording in PR == slick

PR #27, opened 2021-03-01
Add endpoints to retrieve list of subjects

aomi:
Adds `/subjects/{term}` to get a list of subjects with their respective title and shortform.


zthh626 (approved):
Functionality wise, seems good to me.

Tested:
/subjects/{term}

- '202001', '202005', '202009', '202101', '202105'
20201 and 202005 returns `[]` but I think that's course-scraper not finding anything
- 'asdf', '20202222'
Throws `Validate Error`

zthh626:
If we add account functionality, we may need to cache past course data in case users want to see their old data. Although I don't think it matters for the subject endpoint because users can only ever select the current and next 2 terms.

PR #28, opened 2021-03-01
Add sidebar

zthh626:
Branched off feat/course-card-drop-down

Added react-infinite-scroll-component. - Since all the data we need loaded, we don't need to pass a function into the next prop. Same logic with hasMore prop. And since we never call next, we will never see loader prop either.
Sidebar takes prop term: Term.
For some reason couldn't merge master with aomi's useSubject, so I added it myself.
Changed styling for Card and CardDropDown.

@aomi I'm pretty sure I did the data manipulation naively on useGetCourses. Let me know if you want it changed.

**Also, I'll need to add filtering on the courses/subjects when we add the search bar.**

https://user-images.githubusercontent.com/31324580/109458612-fc3fbb80-7a11-11eb-84e7-b749114509dd.mov


package.json
```diff
@@ -19,6 +19,7 @@
     "react": "^17.0.1",
     "react-dom": "^17.0.1",
     "react-icons": "^4.2.0",
+    "react-infinite-scroll-component": "^6.0.0",
```
aomi, line 22: you need to include `package-lock.json`

zthh626:
Styling now looks like this:
Removed extra margin at the bottom of Card.
Removed min width of sidebar.
Stacked text vertically instead of horizontally on CardDropDown.

Screen Shot 2021-02-28 at 11 26 02 PM

src/app/sidebar/Sidebar.tsx
```diff
@@ -1,24 +1,68 @@
...
+  if (courses != null) {
+    for (let i = 0; i < courses.length; i++) {
+      let course: Course = courses[i];
+      let subject = course.subject;
+      if (!(subject in parsedCourses)) {
+        parsedCourses[subject] = [];
+      }
+      parsedCourses[subject].push(course)
+    }
+  }
```
aomi, line 35: Some might argue which is better but.
```ts
  const p = subjects?.reduce(
    (acc, curr) => ({
      ...acc,
      [curr.subject]: courses?.filter((c) => c.subject === curr.subject),
    }),
    {} as { [subject: string]: Course[] | undefined }
  );
  ```
  Not sure which is better give how it turns out.

zthh626: When using filter, it calls for every course. So, I combined reduce and previous logic to make it more efficient.

aomi (approved):

aomi (changes requested):
why do we need both icon libraries?

zthh626:
@aomi I was testing react-icons and forgot to remove it

aomi (approved):

PR #38, opened 2021-03-07
Refactor/sidebar

zthh626:
Removed accordion style sidebar to subjects -> courses directory file. Did this for a number of reasons:

- Accordion style was very unoptimized (slow to load courses)
- Confusing to use, would get lost when scrolling
- Didn't really need the the infinite-scrolling-component because we load all courses in the beginning anyways

Added top bar to display what subject directory it's in. In the future, we should add filtering options (filtering out subjects/courses with no sections available for term).

Removed `CardDropDown` in favour of just using `Card` for both Courses and Subjects.

Added a SidebarContainer that has a loading animation.

Also, thanks @aomi for helping to figure out the optimizations and fixing js prettier!

https://user-images.githubusercontent.com/31324580/110231436-926d5900-7ecc-11eb-81e3-a0e81a9cf9c4.mov


zthh626:
Updated icons for visibility:

Screen Shot 2021-03-07 at 11 12 51 AM
Screen Shot 2021-03-07 at 11 12 56 AM

aomi (approved):

PR #41, opened 2021-03-09
subject alphabetical order

zthh626:


aomi (commented):
Can we sort once so we don't need to sort upon reloading the subjects?

src/app/sidebar/Sidebar.tsx
```diff
@@ -39,24 +39,31 @@ function computeParsedCourses(courses: Course[]) {
...
+  const sortedSubjectCards = useMemo(
+    () =>
+      subjects
+        .sort((a, b) => (a.subject > b.subject ? 1 : b.subject > a.subject ? -1 : 0))
+        .map((subject, index) => (
+          <Box data-subject={subject.subject} onClick={handleSubjectChange} key={index}>
+            <Card subject={subject.subject} title={subject.title} />
+          </Box>
+        )),
+    [subjects]
```
aomi, line 64: I did some reading into the `useMemo` and JSX elements. I think I would prefer we don't memoize the `JSX` elements itself as the expensive computation is the sorting. I believe what happens is if we do it with the elements we end up using more memory and I don't think storing JSX in memory is very beneficial and prefer not doing it.

src/app/sidebar/components/TopBar.tsx
```diff
@@ -14,18 +14,16 @@ export interface TopBarProps {
 
 export function TopBar({ selectedSubject, handleTopBarBackClick }: TopBarProps): JSX.Element {
   return (
-    <HStack bg="white" p="1em" top="0" m="0" boxShadow="md" zIndex={500}>
```
aomi, line 17: We don't want the entire topbar clickable given that's here we're going to likely put features. Also if it is a button, we need to make it more "clickable" anyhow.

aomi (approved):

PR #46, opened 2021-03-11
feat: use Typesaurus to add Typescript capabilities to Firestore

aomi:
The solution to use custom converters etc. didn't support sub collections (afaik) so this adds [a library](https://typesaurus.com/) that does that and a bit more (including handling sub collections). 

In terms of functionality in this PR, it doesn't change the behaviour of the application. This just refactors existing calls to match what is currently running.

It also removes the class encapsulation in section service related code given there isn't a need nor benefit. (this is more of a pre-requsuite to making testing/writing actual tests)


functions/src/sections/Section.service.ts
```diff
@@ -1,123 +1,129 @@
...
-  public async updateSectionMappings(): Promise<void> {
-    // TODO: a good candidate for using RemoteConfig :)
-    const terms = ['202009'];
-    // get all the courses
-    const courses = (await UVicCourseScraper.getCourses()).map((course) =>
-      subjectCodeExtractor(course)
-    );
+export async function updateSectionMappings(): Promise<void> {
+  // TODO: a good candidate for using RemoteConfig :)
+  const terms = ['202009'];
```
keithradford, line 93: I know there's a TODO here but is there a reason this term is hard coded in?

keithradford (approved):
LGTM +1 comment

functions/src/sections/Section.service.ts
aomi: The diff shows this but I didn't do it in this PR.

aomi: nothing calls this endpoint right now (im gonna replace it soon anyhow :shrug: )

PR #70, opened 2021-03-13
Fix maximum depth bug

keithradford:
Fixed a bug where the maximum depth was being exceeded from the `useEffect` call in `TermButtons.tsx`. The variable `terms` was being called within the method which caused an infinite loop.

### DIFF
--- a/src/app/header/components/TermButtons.tsx
+++ b/src/app/header/components/TermButtons.tsx
@@ -4,12 +4,12 @@ import React, { useContext, useEffect, useState } from 'react';
 import { TermContext } from '../../context/TermContext';
 import { getCurrentTerms, getReadableTerm } from '../../shared/utils/terms';
 
+const terms = getCurrentTerms();
+
 export function TermButtons(): JSX.Element {
   const [status, setStatus] = useState([false, false, false]);
 
   const { term, setTerm } = useContext(TermContext);
-
-  const terms = getCurrentTerms();
 
   //initally the current term button needs to be set active to reflect the default term of the context
   useEffect(() => {
@@ -17,7 +17,7 @@ export function TermButtons(): JSX.Element {
     const initStatus = [false, false, false];
     initStatus[idx] = true;
     setStatus(initStatus);
-  }, [term, terms]);
+  }, [term]);
 
   const onClick = (event: React.MouseEvent<HTMLElement, MouseEvent>) => {
     event.preventDefault();


src/app/header/components/TermButtons.tsx
```diff
@@ -4,20 +4,20 @@ import React, { useContext, useEffect, useState } from 'react';
...
 import { getCurrentTerms, getReadableTerm } from '../../shared/utils/terms';
 
+const terms = getCurrentTerms();
+
 export function TermButtons(): JSX.Element {
   const [status, setStatus] = useState([false, false, false]);
 
   const { term, setTerm } = useContext(TermContext);
 
-  const terms = getCurrentTerms();
```
aomi, line 12: Should this be a React hook (or wrapped into a React hook)?

keithradford: Not sure if it really needs to be. To me this is just a global variable being used for reference in the component.

aomi: If you're 100% sure we won't have issues where this being used in different places (and not being in sync) will cause any issues then I think we're fine to leave it as is.

aomi (approved):
If you're sure we won't have issues where `getCurrentTerms` being used in different places (and not being in sync) will not cause any issues then I think we're fine to leave it as is.

src/app/header/components/TermButtons.tsx
```diff
@@ -4,20 +4,20 @@ import React, { useContext, useEffect, useState } from 'react';
 import { TermContext } from '../../context/TermContext';
 import { getCurrentTerms, getReadableTerm } from '../../shared/utils/terms';
 
+const terms = getCurrentTerms();
```
isaaccormack, line 7: Why is this moved to the global scope? Shouldn't just removing `terms` from the deps fix the issue?

keithradford: either works, is there a preferred way?

isaaccormack: Well if it doesn't have to be global it probably shouldn't be

keithradford: I'll just throw the argument out there for it being global: there's a linting rule which wants terms in the dependency array when it's scoped to the function so technically the dependency array will be incomplete. AFAIK both ways fix the bug but both also have "cons".

isaaccormack: ahh ok thanks, global then

isaaccormack (approved):

PR #71, opened 2021-03-14
feat: Add filtering of courses based on availibility 

aomi:
# THIS IS A CHONKER PREPARE YOURSELF
This PR changes the usage of Firestore significantly in order to allow more features in regards to data queries. 

The goal of this PR is to allow courses to be filtered based on whether they are in session for a given term. This allows the application to only show courses that are in session. 

The previous way we were storing course metadata (`course_mappings`, which stored the CRNs for a given course and term ie. `202101SENG265`) 

This did not allow query based on an attribute since the `term` was not stored. The only way to query would be to get all the keys and do some string manipulation on the key itself. That seemed janky so this restructures how data is stored in Firebase.

## Firestore
The key is still constructed the same way where the `term`, `subject`, `code` and `crns` are also stored, however `term` and other fields are added into the document. Section specific data is stored as a subcollection. This results in a document key for a section being `202101SENG265/sections/10001` which would reference a single section document (check the `collections.ts` for details)

## Data Ingest
In order to allow the filtering, we need to have the data on all courses for a given term. Minimally speaking this means the number of sections for a course in a term. This allows the backend to determine if a course is in session (by checking if any CRNs exist for that class). As a result this PR adds a script to populate the database with the required data as a ETL operation. In the future we will need to make the backend periodically refresh (or invalidate) data but for this PR we leave that as future work.

### DIFF
--- a/functions/build/routes.ts
+++ b/functions/build/routes.ts
@@ -18,7 +18,6 @@ const models: TsoaRoute.Models = {
         "properties": {
             "pid": {"dataType":"string","required":true},
             "title": {"dataType":"string","required":true},
-            "dateStart": {"dataType":"string","required":true},
             "subject": {"dataType":"string","required":true},
             "code": {"dataType":"string","required":true},
         },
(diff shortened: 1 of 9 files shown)


keithradford (approved):
LGTM, maybe @isaaccormack could leave some comments too since he's more familiar with firebase but none the less it looks good.

PR #72, opened 2021-03-14
Setup Sentry

aomi:
Adds [Sentry](https://sentry.io/) on the React application and also adds a "release" stage for Sentry integration. This stage adds a "release" but also uploads the source maps. This allows the source maps to be relevant to the error for a given release.


isaaccormack (commented):
looks like this is ok after comparing with docs, would like to test for myself before approving

isaaccormack (approved):

PR #73, opened 2021-03-14
Added basic react-router functionality

shergillmanav19:


src/App.tsx
```diff
@@ -15,25 +15,36 @@ export function App(): JSX.Element | null {
...
+                                information.
+                              </Text>
+                            </VStack>
+                          </Center>
+                        )}
+                      </Flex>
+                    </Flex>
+                  </Box>
+                }
+              />
```
aomi, line 45: @keithradford should we move this code into a `pages` folder?

isaaccormack: agreed, or `routes` folder?

src/App.tsx
```diff
@@ -15,25 +15,36 @@ export function App(): JSX.Element | null {
...
-                      </Text>
-                    </VStack>
-                  </Center>
-                )}
-              </Flex>
-            </Flex>
-          </Box>
+          <Router>
+            <Routes>
+              <Route path="/" />
```
isaaccormack, line 20: shouldn't the `/` route default to the app as we have now? As mentioned in the issue for this, theres no reason to make actual routes right now so I think only having this route, and the code in the route moved into a file in the `pages` folder (as discussed below) should suffice.

isaaccormack (changes requested):

src/App.tsx
shergillmanav19: yep! I could just remove that route, but i think @aomi wanted a "/" route to a blank page and a "/calendar" route to the homepage we have right now

src/App.tsx
keithradford: yeah, I think `routes` makes more sense given they aren't necessarily "pages". Either way I think a directory is justified

shergillmanav19 (commented):
done

src/calendar.tsx
```diff
@@ -0,0 +1,46 @@
...
+
+import { TermContext } from './app/context/TermContext';
+import { Header, Content, SidebarContainer } from './app/index';
+import { getCurrentTerm } from './app/shared/utils/terms';
+import { Term } from './fetchers';
+export function App(): JSX.Element | null {
+  const [term, setTerm] = useState(getCurrentTerm());
+  const [pid, setPid] = useState<string>();
+  const [subject, setSubject] = useState<string>('');
+  const [code, setCode] = useState<string>('');
```
aomi, line 12: This shouldn't have default empty strings.

src/calendar.tsx
```diff
@@ -0,0 +1,46 @@
+import { Box, Center, Flex, Heading, Text, VStack } from '@chakra-ui/react';
+import { useState } from 'react';
+
+import { TermContext } from './app/context/TermContext';
+import { Header, Content, SidebarContainer } from './app/index';
+import { getCurrentTerm } from './app/shared/utils/terms';
+import { Term } from './fetchers';
+export function App(): JSX.Element | null {
```
aomi, line 8: I would refactor this into `Calendar`

aomi (approved):
Once those two issues are resolved LGTM

isaaccormack (approved):

PR #74, opened 2021-03-14
Created bookmark button

shergillmanav19:
Screen Shot 2021-03-13 at 10 09 08 PM

^That is how it looks! I made it a different color so it could pop out.

### DIFF
--- a/src/app/content/components/Course.tsx
+++ b/src/app/content/components/Course.tsx
@@ -1,7 +1,7 @@
 import { Box, Heading, Text } from '@chakra-ui/layout';
-import { BackgroundProps, Divider, Flex } from '@chakra-ui/react';
-import { PropsWithChildren } from 'react';
-
+import { BackgroundProps, Button, Divider, Flex, Spacer } from '@chakra-ui/react';
+import React, { PropsWithChildren } from 'react';
+import { BsFillBookmarkFill } from 'react-icons/bs';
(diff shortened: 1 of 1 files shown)


isaaccormack:
In terms of UI design, IMO it would look better if we moved the bookmark button up on the same level as the header and kept it on the RHS. I think it would look better / still be self explanatory if the button was only the icon, like [this](https://previews.123rf.com/images/sarahdesign/sarahdesign1403/sarahdesign140301142/26990341-bookmark-button.jpg). The button should then be made a bit bigger, like the same height as the title. I do like the style / colour / behaviour of this button though!
I think it would look good then if the credits and other shields were centered on that row.

keithradford:
+1 on @isaaccormack 's comment on the button design. Just having the bookmark icon keeps things simple and better visualizes to a user if a course is bookmarked or not. I.e. if a course isn't bookmarked, it has a white background, then when it's selected it will have some fill to represent that.

isaaccormack:
^ this got me thinking. The button needs to have a hook associated with it to store state so that when the button is pressed, its appearance changes (nothing happens right now).

Also, while outside of the scope of the original issue, I think this PR would be a good place to implement #75 .

shergillmanav19:
sweet ! thanks for the feedback guys will implement that this week!

PR #76, opened 2021-03-17
Add feedback button

### LINKED ISSUE #64: Add a feedback component
When the website is in the hands of the public, we are going to need a way to receive user feedback. We should add some type of pop-up, button, tab, etc. that a user can click to send us feedback. 
A good example of this is https://www.intercom.com/ with the button in the bottom right corner that provides a pop-up. See #63 for what kind of links we would display or get feedback on.

keithradford:
## Overview

Closes #64 

Added a button to the bottom left which adds some options for us to receive user feedback. It provides a simple pop-up with 2 options for the user:

- Google form: You get the gist, a nice way for us to collect feedback and is easy to use
- GitHub discussion: For the more technical side of users, or people with GitHub accounts. Acts as a good platform for any feedback as well as bug reports. Nice to have this information in the repo.

Some notes: Content of the pop-up can change based on feedback. The structure of the whole component is in place so it's easy to change that. Same goes for colour scheme, picked blue for now but can change once we decide site-wide colours. Lastly I changed the width of the content to be a %. Changes how it acts in terms of responsiveness, but shouldn't be noticeable. Only did this so the feedback button wouldn't end up blocking content on smaller windows. 

Screen Shot 2021-03-16 at 5 19 07 PM
Screen Shot 2021-03-16 at 5 18 59 PM


src/App.tsx
```diff
@@ -11,7 +11,7 @@ export function App(): JSX.Element | null {
   const [pid, setPid] = useState<string>();
 
   return (
-    <ChakraProvider>
+    <ChakraProvider portalZIndex={999}>
```
isaaccormack, line 14: why must portalZIndex be set?

aomi:
@keithradford I think we can slim down this UI. Notabily we don't need to show "Google Forms" and it takes up a lot of space. I think Also a smaller GitHub logo would be fine (or maybe something else)... We should put assets into the repository so just download them and throw them into `/public`

src/app/misc/Feedback.tsx
```diff
@@ -0,0 +1,88 @@
...
+        />
+      </PopoverTrigger>
+      <Portal>
+        <PopoverContent pos="sticky" mr="25px" maxH="500px" overflowY="scroll">
+          <PopoverHeader bg="blue.500" color="white" fontWeight="bold" textAlign="center">
+            Feedback
+          </PopoverHeader>
+          <Box px="15px" color="gray.600">
+            <PopoverBody textAlign="left">
+              Clockwork is in the early stages of development and feedback is greatly appreciated! <br />
```
isaaccormack, line 67: Instead this could possible say? 
`Feedback is greatly appreciated during our early stages of development! <br /> Share your thoughts here:`

I like this as:
1. Feedback is the first word in the sentence
2. It is more direct

src/App.tsx
keithradford: Portal z-index is set to 10 on default, it was showing up under elements of content

src/app/misc/Feedback.tsx
```diff
@@ -0,0 +1,88 @@
...
+          <Text>{children}</Text>
+        </Box>
+      </VStack>
+    </Link>
+  );
+}
+
+export function Feedback(): JSX.Element | null {
+  const [isOpen, setIsOpen] = useState(false);
+  const open = () => setIsOpen(!isOpen);
```
isaaccormack, line 41: nifty

src/app/misc/Feedback.tsx
```diff
@@ -0,0 +1,88 @@
...
+
+  return (
+    <Popover closeOnEsc={false} isOpen={isOpen} autoFocus={false}>
```
isaaccormack, line 44: why not closeOnEsc? I feel like the popover should close if someone clicks outside of it also, maybe...?

src/app/misc/Feedback.tsx
```diff
@@ -0,0 +1,88 @@
...
+              Clockwork is in the early stages of development and feedback is greatly appreciated! <br />
+              Share your thoughts with the development team on the following forums:
+              <FeedbackElement
+                src="https://growthsupermarket.com/wp-content/uploads/2018/06/GoogleForms_logo-e1529921391153.png"
+                //TODO: add link once google form is made
+                href=""
+              >
+                Provide your thoughts and feedback on the site by filling out the Google Form.
+              </FeedbackElement>
+              <FeedbackElement
```
isaaccormack, line 76: Bit nitpicky, but could we make the GitHub logo look a little bit nicer like the Google Forms? (ie. looks a bit oversized right now and the logo is being cut off, would be nice if logo and text were same size).

Feel free to omit this

src/App.tsx
isaaccormack: gotcha, thanks

src/app/misc/Feedback.tsx
keithradford: Yep. I'll probably go with the logo-only (not say GitHub and Google Forms) and I can resize images to be exact.

src/app/misc/Feedback.tsx
keithradford: Copied this website pretty closely https://www.intercom.com/
They didn't have closeOnEsc but i'm indifferent

keithradford:
Screen Shot 2021-03-24 at 5 39 54 PM

Cleaned it up from feedback. I don't think text is really needed to go along with the logos.

aomi:
@keithradford Could the width be slightly smaller given there's nothing using the extra space and it seems a bit imbalanced. 
image

keithradford:
@aomi how's this?
Screen Shot 2021-03-25 at 5 29 19 PM

keithradford:
Width on the right is still a bit off, hold on...

keithradford:
Screen Shot 2021-03-25 at 5 37 34 PM

src/app/index.tsx
```diff
@@ -1,3 +1,4 @@
 export { Header } from './header/Header';
 export { Content } from './content/Content';
 export { SidebarContainer } from './sidebar/containers/SidebarContainer';
+export { Feedback } from './misc/Feedback';
```
aomi, line 4: I think it should be in a `feedback` directory imo (misc can get carried away and act as a "catch-all")

aomi (approved):

PR #77, opened 2021-03-17
Add tests for Header

keithradford:
## Overview

Set up Jest and added unit tests for the Header component (got rid of the userbutton component in the header too).


package.json
```diff
@@ -30,14 +29,15 @@
   "scripts": {
     "start": "react-scripts start",
     "build": "react-scripts build",
-    "test": "react-scripts test",
```
aomi, line 33: I'm curious why you needed to do this.

src/app/header/components/__tests__/TermButtons.test.tsx
```diff
@@ -0,0 +1,37 @@
...
+  });
+
+  it('is loaded in with the appropriate button active', () => {
+    render(<TermButtons />);
+
+    // the active button takes time to be rendered in
+    setTimeout(() => {
+      const termButton = screen.getByTestId(term) as HTMLButtonElement;
+      expect(termButton).toHaveAttribute('data-active', '');
+    }, 1000);
```
aomi, line 27: There's ways to avoid this (we defs don't want to slow down tests).

src/app/header/hooks/__tests__/useChangeCallback.test.ts
```diff
@@ -0,0 +1,16 @@
...
+  it('is called with the correct params', () => {
+    const mockOnChange = jest.fn();
+    const event = {
+      preventDefault() {},
+      target: { value: 'test' },
+    };
+    const { result } = renderHook(() => useChangeCallback(mockOnChange));
+    result.current(event as any);
+    expect(mockOnChange).toBeCalledWith(event.target.value);
+  });
```
aomi, line 15: Does this actually test how the hook functions (like passing in deps etc)

src/app/header/components/__tests__/TermButtons.test.tsx
```diff
@@ -0,0 +1,37 @@
+import { fireEvent, render, screen } from '@testing-library/react';
+
+import { getCurrentTerm, getCurrentTerms } from '../../../shared/utils/terms';
+import { TermButtons } from '../TermButtons';
+
+const term = getCurrentTerm();
+const terms = getCurrentTerms();
+
+describe('TermButtons', () => {
+  it('turns the button active on click', () => {
```
isaaccormack, line 10: if the other test below this must wait until the button is loaded to check if the appropriate button is active, does that mean this test is running before the button is fully loaded? Seems like that presents a bit of a race condition?

isaaccormack (approved):

src/app/header/components/__tests__/TermButtons.test.tsx
keithradford: Thoughts on changing 1000 to 1? 😆

package.json
keithradford: Didn't really like how `react-scripts test` ran. IMO Jest was much cleaner as it just runs the tests and that's that.

keithradford:
Closing this because it's super stale and would benefit from a redo.

PR #82, opened 2021-03-23
Basic Search UI implemented with Algolia

### LINKED ISSUE #81: Search UI

aomi:
## Overview

Closes #81 

Adds the search bar functionality and search side bar functionality to the front end app.


keithradford:
Is there overlap with this PR and #71 ? Seems like the first few files have similar changes if I'm recalling correctly so wanted to double check

aomi:
> Is there overlap with this PR and #71 ? Seems like the first few files have similar changes if I'm recalling correctly so wanted to double check

Yeah it rebases off the other PR so. Should have made that clear.

src/App.tsx
```diff
@@ -6,29 +6,38 @@ import { Header, Content, SidebarContainer } from './app/index';
...
-                setCode={setCode}
+                setSelectedCourse={setSelectedCourse}
+                selectedCourse={selectedCourse}
+                searchQuery={query}
               />
               <Flex minW="80%" overflow="auto" justifyContent="center" height="100%" boxShadow="lg" zIndex={56}>
-                {pid && subject && code ? (
-                  <Content term={term as Term} pid={pid} subject={subject} code={code} />
+                {selectedCourse ? (
+                  <Content term={term as Term} selectedCourse={selectedCourse} />
```
keithradford, line 40: Much cleaner 🙌

src/app/content/Content.tsx
```diff
@@ -1,38 +1,21 @@
 import { Box, Skeleton } from '@chakra-ui/react';
 
+import { SelectedCourse } from '../../App';
```
keithradford, line 3: this is being imported a few times, should we make a types.ts?

src/app/header/components/SearchBar.tsx
```diff
@@ -1,29 +1,49 @@
...
+      <InputGroup size="sm">
+        <Input
+          placeholder="Search for courses"
+          bg="white"
+          width={['xs', 'sm', 'md']}
+          value={currentRefinement}
+          onChange={handleChange}
+        />
+        <InputRightAddon p="0">
+          <IconButton aria-label="Search" borderRadius="0" size="sm" icon={<GrSearch />} />
```
keithradford, line 37: this can probably be removed now, think it's redundant

aomi: The icon?

keithradford: Yeah, since pressing it doesn't do anything

src/app/content/Content.tsx
aomi: I'm not sure where as I don't think a `types` file in the root of the application is fitting. Likely we might move stuff out of `App.tsx` into move specific components which might be a good time to make those `types.ts` as needed.

keithradford (approved):

PR #86, opened 2021-03-28
Adding react-big-calendar for scheduler

### LINKED ISSUE #84: Calendar

zthh626:
## Overview
Closes #84 

Uses react-big-calendar. 

**Note: will be adding more styling later, priority was getting calendar component working**
**Note:  Unfortunately [react-big-calendar](https://github.com/jquense/react-big-calendar#localization-and-date-formatting) only uses `Globalize.js`, `date-fns`, or `Moment.js`, so this results in some tech debt since we already use `Dayjs`**

Screen Shot 2021-04-27 at 11 42 18 PM

`ScheduleCalendar` takes a list of the custom interface `CalendarEvent`, which is information about the section + colour of the event to be put on the calendar.

To create recurring events I used RRule, which generates the events using the start date and some specified params, like week, intervals, etc..

### Misc Info

Behaviour of the calendar:
- When no events -> goes to current day
- When there is an event -> goes to min event date

Some nice things to know are in `SchedulerCalendar`:
- `eventStyleGetter` styles the whole event
- `CustomEvent` is the react component that goes within the event that displays all info about the event.

In the future, it is possible for onClicks on events, but that should be a separate ticket.

### Testing to see events

Replace `pages > scheduler > scheduler.tsx` with this
```
import { Flex } from '@chakra-ui/layout';

import { Header, SchedulerContainer } from '../../app';
import { useSections } from '../../shared/fetchers';

export function Scheduler(): JSX.Element {
  const term = '202105';
  const subject = 'ECE';
  const code = '310';
  const { data: scehdulelisting } = useSections({ term, queryParams: { subject, code } });
  return (
    <Flex h="100vh" direction="column">
      <Header />
      {scehdulelisting && (
        <SchedulerContainer
          calendarEvents={(() => {
            const x = new Array();
            x.push({
              subject: subject,
              code: code,
              color: 'green',
              sectionCode: scehdulelisting[0].sectionCode,
              meetingTime: scehdulelisting[0].meetingTimes[0],
            });
            return x;
          })()}
        />
      )}
    </Flex>
  );
}

```


src/app/scheduler/components/SchedulerCalendar.tsx
```diff
@@ -0,0 +1,155 @@
...
+    backgroundColor: resource && resource.color,
+    borderRadius: '0px',
+    opacity: 0.8,
+    color: 'black',
+    border: '0px',
+    display: 'block',
+  };
+  return {
+    style: style,
+  };
```
aomi, line 28: Anything topping us from returning the style object directly?

src/app/scheduler/components/SchedulerCalendar.tsx
```diff
@@ -0,0 +1,155 @@
...
+    }
+
+    return daysRRule;
+  };
+
+  const events = useMemo(() => {
+    const events: Event[] = [];
+    calendarEvents?.forEach((calendarEvent) => {
+      try {
+        if (calendarEvent.meetingTime.time === 'TBA') {
```
aomi, line 77: I think a `indexOf` might be better so if there's some weird formatting it wouldn't die from it.

src/app/scheduler/components/SchedulerCalendar.tsx
```diff
@@ -0,0 +1,155 @@
...
+
+import { CalendarEvent } from './CalendarEvent';
+
+dayjs.extend(utc);
+dayjs.extend(customParseFormat);
+
+const localizer = momentLocalizer(moment);
+
+const eventStyleGetter = ({ resource }: Event) => {
+  var style = {
```
aomi, line 18: If we want to keep a `style` object assignment here, we should have this be a `const`.

src/app/scheduler/components/SchedulerCalendar.tsx
zthh626: This needs to be { className?: string, style?: Object }

src/app/scheduler/components/SchedulerCalendar.tsx
zthh626: I'll change it to a const.

aomi (approved):

PR #95, opened 2021-04-14
Joey/65 header sidebar responsive

J-Raymer:
Responsive changes for non-mobile view.
Additional visual changes such as font size and TopBar layout for review.
Grid layout for header will need to be reverted after mobile view is implemented.


aomi:
@J-Raymer 
I don't think this is an intentional change? (this is full size desktop)
image

J-Raymer:
> @J-Raymer
> I don't think this is an intentional change? (this is full size desktop)
> image

It was for smaller screens I left it up for larger sizes to see if its something we wanted to add. I can change it back for larger screens if you don't like this change.

src/app/content/components/Schedule.tsx
```diff
@@ -28,9 +28,7 @@ export function Schedule({ meetingTimes }: ScheduleProps): JSX.Element {
       <Tbody>
         {meetingTimes.map((m, i) => (
           <Tr key={i}>
-            <Td>
-              <Badge>{m.days}</Badge>
```
keithradford, line 32: The `Badge` import can be removed now (this was the only instance)

src/app/header/Header.tsx
```diff
@@ -32,17 +40,21 @@ export function Header({ onSearchChange }: HeaderProps): JSX.Element {
...
+        colStart={{ base: 1, sm: 1, md: 2 }}
+        colSpan={{ base: 1, sm: 1, md: 1 }}
+        rowStart={{ base: 3, sm: 2, md: 1 }}
+      >
         <Center>
           <Search onChange={onSearchChange} />
         </Center>
       </GridItem>
-      <GridItem colStart={3} alignContent="flex-end">
+      <GridItem colStart={{ base: 1, sm: 2, md: 3 }} rowStart={{ base: 2, sm: 2, md: 1 }} alignContent="flex-end">
```
keithradford, line 57: is there a way you can move the term buttons closer to the search bar? Maybe move it into the same grid item and give it a margin on the left?
IMO it makes sense to be close to the search bar since they're directly related.

src/app/header/components/TermButtons.tsx
```diff
@@ -1,4 +1,4 @@
-import { Button, ButtonGroup } from '@chakra-ui/react';
+import { Button, ButtonGroup, useBreakpointValue } from '@chakra-ui/react';
```
keithradford, line 1: `useBreakpointValue` can be removed here

keithradford (approved):
LGTM

aomi:
@J-Raymer I'm thinking we close this out given we kinda bought ourselves some time in terms of getting mobile working.

PR #107, opened 2021-04-28
Add links to UVic pages at bottom of content

### LINKED ISSUE #106: Add Attribution / Source URL to course / schedule content
Add a source link so people can view where the data is from and verify themselves.
ie. `Source: UVic Undergraduate Calendar 2021` as a link opening a new tab. Ideally we show the page and not the JSON.

keithradford:
## Overview

Closes #106 

Added links to the UVic undergrad calendar and class schedule listings at bottom of content component. The links send the user in a new tab to the courses respective pages for the Kuali and BAN1P data. I've purposely left off the section details link because there would have to be a link for each section and that can be accessed from the listings link.

Screen Shot 2021-04-27 at 7 18 08 PM
Screen Shot 2021-04-27 at 7 27 04 PM

### DIFF
--- a/src/app/content/Content.tsx
+++ b/src/app/content/Content.tsx
@@ -20,7 +20,6 @@ export type ContentProps = {
 export function Content({ term }: ContentProps): JSX.Element {
   const [searchParams] = useSearchParams();
   const { data, loading } = useGetCourse({ term, pid: searchParams.get('pid') || '' });
-  console.log(data);
 
   return (
     <Flex width={['container.md', 'container.lg', 'container.xl']} flexDirection="column" minH="100%">
--- a/src/app/content/Content.tsx
+++ b/src/app/content/Content.tsx
@@ -57,7 +57,7 @@ export function Content({ term }: ContentProps): JSX.Element {
       </Box>
       <Spacer />
       <Center>
-        <Text as="span" fontWeight="bold" fontSize={12}>
+        <Text as="span" fontWeight="bold" fontSize={12} mb={5}>
           Sources:
           <Text as="span" color="blue.500" fontWeight="light">
             <Text
--- a/src/app/content/Content.tsx
+++ b/src/app/content/Content.tsx
@@ -53,8 +53,8 @@ export function Content({ term }: ContentProps): JSX.Element {
           </>
         )}
       </Skeleton>
-      <Center>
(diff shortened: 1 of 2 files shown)


src/app/content/Content.tsx
```diff
@@ -1,57 +1,85 @@
...
+          Sources:
+          <Text as="span" color="blue.500" decoration="underline" fontWeight="light">
+            <Text
+              as="a"
+              href={`https://www.uvic.ca/calendar/undergrad/index.php#/courses/${data?.pid}`}
+              target="_blank"
+              mx="3"
+              _hover={{ color: 'blue' }}
+            >
+              {`UVic Undergraduate Calendar ${new Date().getFullYear()}`}
```
aomi, line 70: Doesn't this need to be tied to the term?

src/app/content/components/Section.tsx
```diff
@@ -127,6 +130,20 @@ export function SectionInfo({
           <SeatInfo seat={seat} />
         </Box>
       </Box>
+      <Text as="span" fontWeight="bold" fontSize={10} align="right" w="100%" display="block">
```
aomi, line 133: Too small. I would make it 12 at least and drop the underline on both.

src/app/content/Content.tsx
keithradford: The link?

aomi: The year. If you select Fall right now it's technically wrong.

keithradford: Ah yes. in that case... the content is static across terms, maybe should just remove the year

aomi: Yeah let's go with that.

src/app/content/Content.tsx
```diff
@@ -1,57 +1,86 @@
...
-}
+};
 
 /**
  * Primary UI component for content
  */
 export function Content({ term }: ContentProps): JSX.Element {
   const [searchParams] = useSearchParams();
   const { data, loading } = useGetCourse({ term, pid: searchParams.get('pid') || '' });
+  console.log(data);
```
aomi, line 24: image

aomi (approved):
delet it

src/app/content/Content.tsx
keithradford: thanks eslint 🙄

PR #113, opened 2021-04-29
Add Schedule SideBars

### LINKED ISSUE #112: Create Right/Left Sidebar for Scheduler
Create the left sidebar, which will be similar to the one currently on the Calendar view of the app but without any of the links working for now as they will behave differently. the ( i ) should link back to the calendar view and the ( + ) button should be there but not tied to any functionality yet.
Left:

The right sidebar should be equal width as the left sidebar. For the time being the right sidebar should only have a topbar and a dummy button not tied to any functionality. 

The finished page should look like this:

zthh626:
## Overview

Closes #112

Adds sidebar to `/scheduler`. 

In this PR, I moved the sidebar code into `shared/SidebarTemplate` because we need to use it in both calendar and scheduler.
This is reflected by adding a `route` prop to sidebar components.
Also, the PR has capability to add a list of cards in to the right list by making a component like `CourseList` or `SubjectsList` from the sidebar components.


src/app/sidebar/components/Card.tsx
```diff
@@ -25,9 +25,47 @@ export interface CardProps {
...
+            size="md"
+            background="null"
+            _hover={{ bg: 'none' }}
+          />
+        </Box>
+      );
+    } else if (code && schedule) {
+      return (
+        <VStack spacing="-0">
+          <IconButton
```
keithradford, line 52: I think the buttons should be consistent with [how Joey styled them](https://discord.com/channels/754490734622802075/798385368096833566/837205375433703484)

src/app/sidebar/components/Card.tsx
```diff
@@ -39,8 +77,8 @@ export function Card({ subject, title, code, selected }: PropsWithChildren<CardP
       my={1}
       cursor="pointer"
```
keithradford, line 78: For scheduler the cursor shouldn't be a pointer since clicking the card and not the buttons doesn't do anything

src/app/sidebar/components/CoursesList.tsx
```diff
@@ -8,25 +8,34 @@ import { Card } from './Card';
 
 type CoursesListProps = {
   term: string;
+  route: string;
```
keithradford, line 11: Instead of passing route as a prop, you can do something like [this](https://github.com/VikeLabs/clockwork/blob/main/src/app/header/components/NavButtons.tsx#L6)

src/app/sidebar/components/SubjectsList.tsx
```diff
@@ -7,11 +7,23 @@ import { KualiSubject } from '../../../shared/fetchers';
...
   subjects: KualiSubject[];
+  /**
+   * Route that the subject clicked will go to
+   * EX) calendar or scheduler
+   */
+  route: String;
 };
 
-export function SubjectsList({ term, subjects }: SubjectsListProps): JSX.Element {
+export function SubjectsList({ term, subjects, route }: SubjectsListProps): JSX.Element {
```
keithradford, line 26: same thing about route

keithradford: this happens a few more times just didn't add comments

src/shared/SidebarTemplate.tsx
```diff
@@ -0,0 +1,56 @@
...
+import { Term } from './fetchers';
+import { useSessionStorage } from './hooks/storage/useSessionStorage';
+
+export interface SidebarTemplateProps {
+  title: string;
+  children: JSX.Element;
+  term: Term;
+}
+
+export function SidebarTemplate({ children, term, title }: SidebarTemplateProps): JSX.Element {
```
aomi, line 19: I'd be more favoured to use `PropsWithChildren` so we don't have to pass in the `child` as a prop but rather a child node.
https://www.newline.co/@bespoyasov/how-to-define-props-with-children-in-react-typescript-app--56bd18be

src/app/sidebar/components/CoursesList.tsx
```diff
@@ -15,18 +15,28 @@ type CoursesListProps = {
...
+        </LinkBox>
+      );
+    else if (scheduleMatch) {
+      return <Card title={title} subject={subject} code={code} schedule={true} />;
+    }
+  };
 
   if (!courses || !courses[subject]) {
     return null;
   }
```
aomi, line 35: This is SUPER nitty (and I don't exactly know if it is the case) but moving this above your `createCard` is likely a good idea. Since `createCard` is defined each time `CoursesList` is rendered, it is being defined when not being using which in this case is the `!courses || !courses[subject]`.

aomi (approved):

PR #116, opened 2021-04-30
Manav/feat/bookmark storage

### LINKED ISSUE #75: Bookmarking course should save course to local storage
When a user bookmarks a course, the course should be saved to [local browser storage](https://www.robinwieruch.de/local-storage-react) so the bookmarked courses can be persisted between browser sessions, without the user needing an account. It would be nice if there was a button to clear all bookmarked courses also developed for this issue. There should be an option to hide this button on the UI so we can turn it off in the staging env. Maybe the button would just go in the navbar for now.

Off the top of my head, a simple way to implement this would be to store an array of bookmarked courses in local storage. Each course in this array can be indexed on its term, subject, and code.

shergillmanav19:
## Overview

Closes #75


src/shared/hooks/storage/useSavedCourses.ts
```diff
@@ -0,0 +1,50 @@
...
+  deleteCourse: (pid: string, term: string) => void;
+  clearCourses: () => void;
+  contains: (pid: string, term: string) => boolean;
+};
+
+export const useSavedCourses = (): SavedCourses => {
+  const [data, setData] = useLocalStorage<Course[]>('user:saved_courses', []);
+
+  const contains = (pid: string, term: string): boolean => {
+    const result = data.find((course) => course.pid === pid && course.term === term);
```
aomi, line 22: You can reduce this to `return !!data.find`

src/shared/hooks/storage/useSavedCourses.ts
```diff
@@ -0,0 +1,53 @@
...
+  subject: string;
+  code: string;
+  pid: string;
+  term: string;
+};
+
+type SavedCourses = {
+  courses: Course[];
+  addCourse: (newCourse: Course) => void;
+  deleteCourse: (pid: string, term: string) => void;
```
aomi, line 13: These should have the same args.

src/shared/hooks/storage/useSavedCourses.ts
```diff
@@ -0,0 +1,53 @@
...
+
+  const deleteCourse = (pid: string, term: string): void => {
+    // find the course, delete if found
+    const newArr: Course[] = data.filter((course) => {
+      if (course.pid === pid) {
+        return course.term !== term;
+      }
+      return course.pid !== pid;
+    });
+    setData(newArr);
```
aomi, line 45: This can be simplified greatly. Not sure why it's not just a compound equality check for `pid` and `term`?

src/app/content/Content.tsx
```diff
@@ -22,6 +27,18 @@ export function Content({ term }: ContentProps): JSX.Element {
   const [searchParams] = useSearchParams();
   const { data, loading } = useGetCourse({ term, pid: searchParams.get('pid') || '' });
 
+  const { courses, addCourse, deleteCourse, contains } = useSavedCourses();
+
+  const courseIsSaved = contains(data?.pid!, term);
+
+  const handleBookmarkClick = useCallback(() => {
+    if (!courseIsSaved) {
```
aomi, line 35: If you check if `data` exists in here you can avoid all the checks inside and not have to do the `?` everywhere.

src/shared/hooks/storage/useSavedCourses.ts
```diff
@@ -0,0 +1,53 @@
+import useLocalStorage from './useLocalStorage';
```
aomi, line 1: I think it would make sense to leave this out of the `storage` folder to imply that the other hooks (sessionStorage/localStorage) are hooks to be used to build out other hooks like this one.

aomi (commented):
Need to clean up some unused imports too :wink:

aomi (approved):
Rename `useStorage.tsx` (the file with `useSessionStorage` so it's clear from the filename.

PR #118, opened 2021-05-01
Add landing page

### LINKED ISSUE #115: Implement Landing Page
Implement the landing page where the current "Explore Courses" is displayed.
WIP

keithradford:
## Overview

Closes #115 

Added a landing page. Shows up at the homepage and default calendar page. Pretty straight forward, copied the figma mock-up agreed upon at the meeting.

Screen Shot 2021-05-01 at 2 42 49 PM
Screen Shot 2021-05-01 at 2 42 46 PM


src/app/landing/index.tsx
```diff
@@ -0,0 +1,45 @@
...
+import { Flex, Grid, GridItem, Heading, Text } from '@chakra-ui/layout';
+import { useMediaQuery } from '@chakra-ui/media-query';
+
+export default function Landing() {
+  const [isLargerThan1100] = useMediaQuery('(min-width: 1100px)');
+  console.log(isLargerThan1100);
+
+  return (
+    <Grid
+      templateColumns={isLargerThan1100 ? 'repeat(2, 1fr)' : 'repeat(1, 1fr)'}
```
aomi, line 11: Is there anything stopping using the `xl` breakpoint value to accomplish this? 
`templateColumns={{xl: 'repeat(2, 1fr)', default: 'repeat(1, 1fr)'}}}`? (or something like that)

keithradford: only problem is using breakpoints makes it inconsistent with https://github.com/VikeLabs/clockwork/pull/118/files#diff-5556e38476f1d1d808a8977a1b10585e696de85182714f6f9fa62108f1cd1565R31, so the computer image will go under the text for a bit, and then disappear. this way does both at the exact same width

keithradford: both as in going to 1 column and getting rid of image

src/app/landing/index.tsx
```diff
@@ -0,0 +1,45 @@
...
+        <Text fontSize="25.71">CourseUp makes it simple to browse and schedule UVic Courses</Text>
+        <Flex alignItems="center" mt="40px" as="a" w="fit-content" target="_blank" href="https://www.vikelabs.ca">
+          <Text color="#4C6EA5" fontSize="24.25" mr="5px">
+            Built by students @
+          </Text>
+          <Image src={process.env.PUBLIC_URL + '/assets/vikelabs.svg'} alt="VikeLabs" />
+        </Flex>
+      </GridItem>
+      {isLargerThan1100 && (
+        <GridItem maxH="563px" minW="fit-content">
```
aomi, line 32: You could do `display: none`?

keithradford: wdym?

keithradford: I get `display: none` but curious why

aomi: My perceived usage of `useBreakpoint` was moreso for things JS like maybe changing the behaviour of how a call is made or something and not for stuff like this is my take on it. Also the `sm, xl`, etc. breakpoints are consistent given they are globally set on the theme.

aomi: Performance wise I think it's a bit faster since I _think_ there's only an initial runtime cost but with this there is a JS cost each time (not significant for this but speaking generally)

keithradford: I might be missing your idea on the implementation, but I don't think this is working 😕

aomi: https://chakra-ui.com/docs/features/responsive-styles#the-object-syntax with the keyword being `upwards`.

keithradford: `upwards` 😡

src/app/landing/index.tsx
```diff
@@ -0,0 +1,39 @@
...
+          <Image src={process.env.PUBLIC_URL + '/assets/vikelabs.svg'} alt="VikeLabs" />
+        </Flex>
+      </GridItem>
+      <GridItem maxH="563px" minW="fit-content" display={{ base: 'none', xl: 'initial' }}>
+        <Image
+          src={process.env.PUBLIC_URL + '/assets/computer.svg'}
+          flex="1"
+          style={{
+            WebkitFilter: 'drop-shadow( 27px 8px 36px rgba(0, 0, 0, .25))',
+            filter: 'drop-shadow( 27px 8px 36px rgba(0, 0, 0, .25))',
```
aomi, line 33: I think https://chakra-ui.com/docs/features/the-sx-prop might clean this up.

aomi (approved):
Try out the `sx` prop otherwise LGTM.

PR #128, opened 2021-05-06
CourseCard

### LINKED ISSUE #102: Implement Course Card with Checkbox
[Figma](https://www.figma.com/file/Lvep4lP1867IZUiVQHUtaz/CourseWorks-Clockwork?node-id=211%3A9207)

Please check the Figma and set the scale to 100% in order to get the right scaling. The card heading size and text size should be the same as the existing `Card` component. This component should implement a check-box props (selected, onchange etc) and `onDelete`, `onInfo` callback props that correlate to the button respectively.

The delete and add button should be implemented as shared components and implemented using the Chakra button as a base.

J-Raymer:
Closes #102 

Course Card component for scheduler.


src/app/scheduler/components/CourseCard.tsx
```diff
@@ -0,0 +1,72 @@
...
+    code: string;
+    colour: string;
+    pid: string;
+    selected: boolean;
+    handleChange: Function;
+    handleInfo: Function;
+}
+
+export function CourseCard({ term, subject, title, code, colour, pid, handleChange, handleInfo, selected }: CourseCardProps): JSX.Element {
+    const { data } = useGetCourse({ term, pid: pid || '' })
```
keithradford, line 22: There's no need to use this function if you're passing in everything `data` offers as a prop. I'd pass `term` and `pid` as props, then get the rest of the information with this function fall. That'll clean up the props.

src/app/scheduler/components/CourseCard.tsx
```diff
@@ -0,0 +1,72 @@
...
+                        <Text fontSize="lg" fontWeight="bold">
+                            {subject} {code}
+                        </Text>
+                        <Text fontSize="sm" fontWeight="normal">
+                            {title}
+                        </Text>
+                    </VStack>
+                    <VStack alignContent="right" pr="5px" py="5px">
+                        <IconButton aria-label="Add to Scheduler" icon={!courseIsSaved ? <AddIcon color="white" /> : <CloseIcon color="white" />} background={!courseIsSaved ? "green.400" : "red"} size="sm" onClick={onCourseClick} />
+                        <IconButton aria-label="More information" icon={<InfoOutlineIcon color="white" />} size="sm" background="blue.400" onClick={onInfo} />
```
keithradford, line 66:
```suggestion
                        <IconButton aria-label="More information" icon={<InfoOutlineIcon color="white" />} size="sm" background="blue.400" as={Link} to={`/calendar/${term}/${subject}?pid=${pid}`} />
```
Where `Link` is imported from `react-router-dom`. Removes the need for an `onClick` function.

src/app/scheduler/components/CourseCard.tsx
```diff
@@ -0,0 +1,72 @@
...
+                deleteCourse({ subject: data.subject, code: data.code, pid: data.pid, term })
+            }
+        }
+    }, [data, courseIsSaved, addCourse, term, deleteCourse])
+
+    const onInfo = () => {
+        handleInfo()
+    }
+
+    const onChange = () => {
```
keithradford, line 40:
```suggestion
    const onChange = useCallback(() => {
    handleChange();
    }, [handleChange])
```

aomi:
@J-Raymer make you hit "Resolve conversation" on anything you've addressed/fixed.

src/app/scheduler/components/CourseCard.tsx
```diff
@@ -0,0 +1,92 @@
...
+
+import { Term } from '../../../shared/fetchers';
+import { useSavedCourses } from '../../../shared/hooks/useSavedCourses';
+
+export type CourseCardProps = {
+  term: Term;
+  subject: string;
+  title: string;
+  code: string;
+  colour: string;
```
aomi, line 13: This should be `color?: BackgroundProps['bg']`. Sadly the American spelling just to be consistent :\

src/app/scheduler/components/CourseCard.tsx
```diff
@@ -0,0 +1,92 @@
...
+
+  const onCourseClick = useCallback(() => {
+    if (term && pid !== '' && code !== '') {
+      if (!courseIsSaved) {
+        addCourse({ subject, code, pid, term });
+      } else {
+        deleteCourse({ subject, code, pid, term });
+      }
+    }
+  }, [term, pid, code, courseIsSaved, addCourse, subject, deleteCourse]);
```
aomi, line 40: I think this can be removed and the component using `CourseCard` can be responsible for this stuff.

src/app/scheduler/components/CourseCard.tsx
```diff
@@ -0,0 +1,92 @@
...
+            <Text fontSize="lg" fontWeight="bold">
+              {subject} {code}
+            </Text>
+            <Text fontSize="sm" fontWeight="normal">
+              {title}
+            </Text>
+          </VStack>
+          <VStack alignContent="right" pr="5px" py="5px">
+            <IconButton
+              aria-label="Add to Scheduler"
```
aomi, line 73: :clap: good practice :clap:

src/app/scheduler/components/CourseCard.tsx
```diff
@@ -0,0 +1,92 @@
...
+              background={!courseIsSaved ? 'green.400' : 'red'}
+              size="sm"
+              onClick={onCourseClick}
+            />
+            <IconButton
+              aria-label="More information"
+              icon={<InfoOutlineIcon color="white" />}
+              size="sm"
+              background="blue.400"
+              as={Link}
```
aomi, line 84: This is the wrong `Link`. You want the `Link` from `react-router`.

src/app/scheduler/components/CourseCard.tsx
```diff
@@ -0,0 +1,92 @@
...
+            </Text>
+            <Text fontSize="sm" fontWeight="normal">
+              {title}
+            </Text>
+          </VStack>
+          <VStack alignContent="right" pr="5px" py="5px">
+            <IconButton
+              aria-label="Add to Scheduler"
+              icon={!courseIsSaved ? <AddIcon color="white" /> : <CloseIcon color="white" />}
+              background={!courseIsSaved ? 'green.400' : 'red'}
```
aomi, line 75: I'd argue we shouldn't inverse `courseIsSaved` when we can swap the other of the conditional.

src/app/scheduler/components/CourseCard.tsx
```diff
@@ -0,0 +1,92 @@
...
+        <Flex background={colour} alignItems="center" justifyContent="center" mr="10px">
+          <Flex>
+            <Checkbox
+              backgroundColor="whiteAlpha.600"
+              colorScheme="whiteAlpha"
+              iconColor="black"
+              size="lg"
+              mx="7px"
+              isChecked={selected}
+              onChange={onChange}
```
aomi, line 58: We want the entire card to be a clickable select (not just the checkmark). I think you can make the outermost `Box` be `as={"label"} which should make the entire thing the clickable thing?

src/app/scheduler/components/CourseCard.tsx
```diff
@@ -0,0 +1,92 @@
...
+    }
+  }, [term, pid, code, courseIsSaved, addCourse, subject, deleteCourse]);
+
+  const onChange = useCallback(() => {
+    handleChange();
+  }, [handleChange]);
+
+  return (
+    <Box boxShadow="md" cursor="pointer">
+      <Flex direction="row" background={!selected ? 'blackAlpha.200' : 'white'}>
```
aomi, line 48: Same thing here about the unnecessary negation.

src/app/scheduler/components/CourseCard.tsx
```diff
@@ -0,0 +1,89 @@
...
+export type CourseCardProps = {
+  term: Term;
+  subject: string;
+  title: string;
+  code: string;
+  color: BackgroundProps['bg'];
+  pid: string;
+  selected: boolean;
+  handleChange: Function;
+  handleBookmarkClick: Function;
```
aomi, line 18: These should be `onCourseChange` and maybe `onCourseDelete` where we use `handle` prefix inside components.
Ideally this component should act like a `checkbox` itself.

src/app/scheduler/components/CourseCard.tsx
```diff
@@ -0,0 +1,89 @@
...
+  title,
+  code,
+  color,
+  pid,
+  selected,
+  handleChange,
+  handleBookmarkClick,
+}: CourseCardProps): JSX.Element {
+  const { contains } = useSavedCourses();
+  const courseIsSaved = contains(pid, term);
```
aomi, line 33: This should be removed as well. The reason being the same as the other removals of hooks.

src/app/scheduler/components/CourseCard.tsx
```diff
@@ -0,0 +1,89 @@
...
+            </Text>
+            <Text fontSize="sm" fontWeight="normal">
+              {title}
+            </Text>
+          </VStack>
+          <VStack alignContent="right" pr="5px" py="5px">
+            <IconButton
+              aria-label="Add to Scheduler"
+              icon={courseIsSaved ? <CloseIcon color="white" /> : <AddIcon color="white" />}
+              background={courseIsSaved ? 'red' : 'green.400'}
```
aomi, line 72: I believe when this component will be rendered, it will be saved so we don't run into the issue of "Adding" and are only concerned with removing.

keithradford (approved):
LGTM, [`Section.tsx` ](src/app/content/components/Section.tsx) has some unused imports that need to be removed but good to go once those are done 👍

PR #129, opened 2021-05-07
Add content page current state context

aomi:
## Overview

Closes #XX

- Changes the button "Bookmark" to "Add to Timetable" and "Remove from Timetable" as bookmarking can be confusing on whether it's a browser bookmark. Could simplify the wording but for now it's much more clear to the user.
- Adds skeleton loading to the course content title / description and links. Before it allowed `undefined` links which would be broken. 
- Disables the add/delete button while data is loading. 
- Moves the CRN to the left given it's not very important.
- Course detail badges are smaller now.  
- Adds toast messages when course is added/deleted via the button on the course details view.
- When getting course details errors out, shows the error message (shown below)

image

### DIFF
--- a/src/app/content/components/Section.tsx
+++ b/src/app/content/components/Section.tsx
@@ -72,42 +72,44 @@ export function SectionInfo({
 
   return (
     <Box as="section" bg="white" color="black" my="4" boxShadow="md" p="3" rounded="lg">
-      <Flex my="2" alignItems="center">
-        <Heading mr="5" size="lg" as="h2" whiteSpace="pre">
-          {sectionCode}
-        </Heading>
-        <Heading size="lg" as="h3" color="gray">
+      <Flex my="2" alignItems="center" justifyContent="space-between">
+        <Flex alignItems="center">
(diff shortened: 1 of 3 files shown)


src/app/content/components/Section.tsx
```diff
@@ -72,42 +72,44 @@ export function SectionInfo({
...
+            <Popover>
+              <PopoverTrigger>
+                <Badge colorScheme="green" mx="1">
+                  {instructionalMethod}
+                </Badge>
+              </PopoverTrigger>
+              <PopoverContent>
+                <PopoverArrow />
+                <PopoverCloseButton />
+                <PopoverBody>This means this section takes place online.</PopoverBody>
```
keithradford, line 90: I know this is unrelated to this PR, but if `instructionalMethod` is in-person, does the popover body still say `This means this section takes place online.`?

keithradford (approved):
LGTM, just a side note

src/app/content/components/Section.tsx
aomi: No :disappointed:

PR #130, opened 2021-05-07
add section table(radio) for scheduler

### LINKED ISSUE #114: Implement Sections Table/Radio Select Component
[Figma](https://www.figma.com/file/Lvep4lP1867IZUiVQHUtaz/CourseWorks-Clockwork?node-id=211%3A9625)

Each section should be a `<Option>` but the component `<Sections/>` to be implemented should take in a list of sections that will render out the sections in a table. The table should be encapulated as a `<Radio>` and each section should be an `<option/>`. This `<Sections/>`  component should have a `onChange` that is linked to the `onChange` of the `<Radio/>`

Use https://chakra-ui.com/docs/form/radio as a example.

chsunwoo1002:
## Overview

Closes #114 

add section table as scheduler component
Please let me know if I need to fix or modify it :)

Screen Shot 2021-05-06 at 10 58 48 PM

### DIFF
--- a/src/app/scheduler/components/SchedulerSections.tsx
+++ b/src/app/scheduler/components/SchedulerSections.tsx
@@ -5,7 +5,6 @@ import { ClassScheduleListing, MeetingTimes } from '../../../shared/fetchers';
 
 export function Sectionss({ sections }: { sections: ClassScheduleListing[] }): JSX.Element {
   const [section, setSection] = React.useState('');
-  console.log(section);
   return (
     <RadioGroup onChange={setSection} value={section}>
       <Table size="sm">
--- a/src/app/scheduler/components/SchedulerSections.tsx
+++ b/src/app/scheduler/components/SchedulerSections.tsx
@@ -1,7 +1,7 @@
 import { Radio, RadioGroup, Thead, Table, Tbody, Tr, Th, Td } from '@chakra-ui/react';
 import React from 'react';
 
-import { ClassScheduleListing, MeetingTimes, Sections } from '../../../shared/fetchers';
+import { ClassScheduleListing, MeetingTimes } from '../../../shared/fetchers';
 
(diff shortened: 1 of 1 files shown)


aomi:
@chsunwoo1002 You'll want to use https://chakra-ui.com/docs/data-display/table as the base or else as you can see, the alignment breaks between each section. In your screenshot the "TBA" isn't vertically aligned.

chsunwoo1002:
> @chsunwoo1002 You'll want to use https://chakra-ui.com/docs/data-display/table as the base or else as you can see, the alignment breaks between each section. In your screenshot the "TBA" isn't vertically aligned.

Im gonna change the component as table to make vertically align.
the last component is about the location(where), is that correct?

src/app/scheduler/components/SchedulerSections.tsx
```diff
@@ -0,0 +1,59 @@
...
+            <Th>Section</Th>
+            <Th>Days</Th>
+            <Th>Times</Th>
+            <Th>Location</Th>
+          </Tr>
+        </Thead>
+        <Tbody>
+          {sections.map(({ sectionCode, meetingTimes }) => (
+            <Option sectionCode={sectionCode} meetingTimes={meetingTimes} />
+          ))}
```
aomi, line 23: So this is where the complexity lies... 
If you think about it, you should only be able to select 1 section / section type. This means we obviously can't select two lecture sessions (ie you don't register for `A01` and `A02`). But if there are tutorials and labs, we want to select at least one of them. I don't think this is going to enforce that rule.

aomi:
Thanks for providing screenshots :)

keithradford (approved):
LGTM. There's some imports that need to be removed from [`Sections.tsx`](https://github.com/VikeLabs/clockwork/pull/130/files#file-src-app-content-components-section-tsx-L7), once those are dealt with good to go!

PR #170, opened 2021-06-02
feat: Add Basic Open Graph Preview 

aomi:
Open Graph makes the link when shared on social media nice with a preview etc. This is a basic implementation until we have more capacity for this.

cc: @Brycce


Brycce:
There was a version of it that was working, if you even just go back through the commits and test each one you'll probably find the one that works. I won't have time for it today.

aomi:
> There was a version of it that was working, if you even just go back through the commits and test each one you'll probably find the one that works. I won't have time for it today.

The reason it doesn't work is that the resource doesn't exist. Assume that it does and point the image to `content="https://courseup.vikelabs.dev/assets/laptop.png"` and so forth. 

Fix the typo (`CourseUp`) as the title and this should be fine to merge. When it deploys the assets will be uploaded.

keithradford (approved):

PR #179, opened 2021-06-05
add issue templates

keithradford:


.github/ISSUE_TEMPLATE/feature_request.md
```diff
@@ -0,0 +1,17 @@
...
+---
+
+**Is your feature request related to a problem? Please describe.**
+<!--- A clear and concise description of what the problem is. Ex. I'm always frustrated when [...] -->
+
+**Describe the solution you'd like**
+<!--- A clear and concise description of what you want to happen. -->
+
+**Additional context**
+<!--- Add any other context or screenshots about the feature request here. -->
```
aomi, line 17: I'm hesitant to allow for feature requests (or make it seem like we're wide open to them) till we get that figure out. 
ie. I feel if you want a feature, you need to justify it too like:
"This feature would be good as it would help solve X, Y, Z" and avoid feature requests that don't justify themselves?

aomi (approved):

PR #190, opened 2021-06-06
Registration view

### LINKED ISSUE #162: Implement registration view
See mock-up #160

Implement the UI and components for the registration view. Should be able to implement based off the mock-up :)

keithradford:
## Overview

Closes #162 

Adds a registration page users can access via the timetable which provides a link to UVic's registration page & easy-to-copy CRNs for each of their saved sections. The idea is users can quickly copy and paste CRNs into UVic's registration page to easily register based on their timetable made on the website.

**looking for any a11y recommendations** 

Screen Shot 2021-06-06 at 2 28 00 PM
Screen Shot 2021-06-06 at 2 28 37 PM
Screen Shot 2021-06-06 at 2 28 19 PM
Screen Shot 2021-06-06 at 2 28 09 PM

### DIFF
--- a/src/app/registration/components/RegistrationSection.tsx
+++ b/src/app/registration/components/RegistrationSection.tsx
@@ -1,7 +1,7 @@
 import { IconButton } from '@chakra-ui/button';
 import { Checkbox } from '@chakra-ui/checkbox';
 import { useClipboard } from '@chakra-ui/hooks';
-import { Box, Heading, HStack, Text } from '@chakra-ui/layout';
+import { Box, Heading, HStack } from '@chakra-ui/layout';
 import { useToast } from '@chakra-ui/toast';
 import { useCallback, useEffect } from 'react';
 import { IoCopyOutline } from 'react-icons/io5';
@@ -33,8 +33,13 @@ export function RegistrationSection({ section, crn, seats, selected, handleChang
     <Box py="2">
       <HStack justifyContent="space-between">
         <Heading size="md" as="h3" mt="2">
-          {section}{' '}
-          <Text as="span" fontWeight="normal" ml="2" color="rgb(155, 155, 155)">
+          {section}
+        </Heading>
+        <HStack spacing="3">
+          <Checkbox isChecked={selected} onChange={onChange}>
+            Registered
+          </Checkbox>
+          <Heading size="md" as="h3" color="gray">
             {crn}
             <IconButton
               icon={<IoCopyOutline />}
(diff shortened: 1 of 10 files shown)


aomi:
@keithradford looking sweet so far. Some initial thoughts.
- Let's use an ordered list for the steps mentioned in the top paragraph.
- The `title` of the page should use the "readable" representation.
- The CRN text styling should match the one on the calendar view.
  - I think matching where the CRN is too would be nice. Perhaps moving the select checkbox left of the CRN might work?
- The sections "A01, B03" etc. should match the CRN text styling like the CRN the calendar I think.
- Tightening the margins/padding overall I think would slim the design down a bit.

Extras:
- I think showing the "additional notes" and "meeting times" for the selected sections would be beneficial as it often contains registration specific information.
- Showing the title of the course I think next to the subject/code would be a nice touch.
- When it collapses, it briefly shows two subject/code which is a bit weird but not a huge deal for now imo.
- This page should be pretty easy to get responsive so I think getting it working would be a nice improvement.

src/app/registration/components/RegistrationSection.tsx
```diff
@@ -0,0 +1,61 @@
...
+            {crn}
+            <IconButton
+              icon={<IoCopyOutline />}
+              onClick={onCopy}
+              aria-label="copy"
+              colorScheme="white"
+              color="black"
+              _focus={{
+                outline: 'none',
+              }}
```
FlyteWizard, line 47:
```suggestion
              _focus={{
                outline: 'default',
              }}
```

It is not recommended to remove focus on elements. When a user uses the keyboard to navigate they will be uncertain what is focus if the focus is removed. I think keeping the default focus, or styling the focus is the best option. 

You can read more here: https://webaim.org/techniques/keyboard/

keithradford: thoughts on this solution for a11y? Having the border doesn't look great which is why I removed it so I'm trying some alternatives
Screen Shot 2021-06-06 at 3 53 22 PM

aomi (commented):
Vertical alignment isn't there as well as just width not responding. I think it's because of a usage of `width` and not `min/max Width`? 
image

src/app/registration/components/RegistrationSection.tsx
FlyteWizard: I don't think that is enough of a difference for users. The hard thing is is that aesthetic doesn't override accessibility. 

Can you make the icon bigger or more thick when focused? Ideally for convention, developers should leave the outline on focus. Can you try making the outline a little smaller so it doesn't overlap with the CRN?

keithradford: Making it bigger on focus looks nice. I know I'm jumpng hurdles around not having the outline, but it's tough to make it look nice. And yeah, 100% accessibility over aesthetic but if there's a perfect balance between the 2 that's ideal. Of course if there isn't, the outline will be fine.

aomi (approved):

PR #191, opened 2021-06-06
gave minW and W to box components

### LINKED ISSUE #154: Fix alignment of columns between sections of a course.
The alignment between sections within a course is broken as shown below. 

The columns should align between sections within a course but also between courses. In the above image this would mean the days (ie. `TWF, `MR`) etc. are vertically aligned.

shergillmanav19:
## Overview

Closes #154


aomi (approved):
image

PR #192, opened 2021-06-06
Add ghost events

### LINKED ISSUE #189: Add ghost events
Right now when you're in the timetable view, if an event isn't happening during your current week view, you won't see it at all. We shouldn't assume people will know to move around to different weeks to see if labs are in session and changes not having visible effects can be confusing/appear broken to users.

A solution to this would be to display "ghost" events every day an event occurs during weeks it isn't present. I.e. if a lab starts in the week of September 14th and is Wednesday 2:30-4:30, any week prior will still show it during these times, but it will look like this:

zthh626:
## Overview

Closes #189

### DIFF
--- a/src/app/scheduler/components/SchedulerCalendar.tsx
+++ b/src/app/scheduler/components/SchedulerCalendar.tsx
@@ -6,10 +6,6 @@ import getDay from 'date-fns/getDay';
 import * as enUS from 'date-fns/locale';
 import parse from 'date-fns/parse';
 import startOfWeek from 'date-fns/startOfWeek';
-import dayjs from 'dayjs';
-import customParseFormat from 'dayjs/plugin/customParseFormat';
-import utc from 'dayjs/plugin/utc';
-import _ from 'lodash';
 import { MutableRefObject, useCallback, useMemo, useRef, useState } from 'react';
 import 'react-big-calendar/lib/sass/styles.scss';
 import '../../shared/styles/CalendarStyles.scss';
@@ -18,9 +14,6 @@ import { useParams } from 'react-router';
 import { RRule } from 'rrule';
 
 import { CalendarEvent } from '../shared/types';
-
-dayjs.extend(utc);
-dayjs.extend(customParseFormat);
 
 const locales = {
   'en-US': enUS,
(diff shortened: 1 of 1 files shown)


src/app/scheduler/components/SchedulerCalendar.tsx
```diff
@@ -109,6 +110,15 @@ export function SchedulerCalendar({ calendarEvents }: SchedulerCalendarProps): J
     return today;
   }, [term]);
 
+  const getTermMonthLowerBound = useCallback(() => {
+    const month = /\d{4}(\d{2})/.exec(term);
+    const year = /(\d{4})\d{2}/.exec(term);
+
+    const lowerBound = dayjs.utc(`${month ? month[1] : '09'}, 1, ${year ? year[1] : 2021}`, 'MM, D, YYYY');
```
keithradford, line 117: instead of defaulting to `09` lets default to `01`. I know it's pretty much impossible for it to happen, with `09` as the default, if it's the summer or winter term, the lower bound will be after the term ends.

keithradford: can't add a comment there but on line 258 you're missing `getTermMonthLowerBound` from the dependency array. If adding it breaks the program, you can add an eslint ignore so the linter doesn't get mad

keithradford (approved):

aomi (changes requested):
This is pretty critical. Putting a hold till we QA further. My first thought is what happens where there's labs biweekly etc.

src/app/scheduler/components/SchedulerCalendar.tsx
```diff
@@ -181,14 +192,17 @@ export function SchedulerCalendar({ calendarEvents }: SchedulerCalendarProps): J
...
-          .toDate();
-        const startLowerDateRRule = dayjs
-          .utc(startEndDates[0].trim() + startEndHours[1].trim(), 'MMM, D, YYYY h:mm a')
-          .toDate();
+
+        const courseStartDate = dayjs.utc(startEndDates[0].trim(), 'MMM, D, YYYY').toDate();
+
+        const startDateString = `${dayjsLowerBound.month() + 1}, 1, ${dayjsLowerBound.year()} `;
+        const startUpperDateRRule = dayjs.utc(startDateString + startEndHours[0].trim(), 'M, D, YYYY h:mm a').toDate();
+        const startLowerDateRRule = dayjs.utc(startDateString + startEndHours[1].trim(), 'M, D, YYYY h:mm a').toDate();
```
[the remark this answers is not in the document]
zthh626: Oops, I forgot that we said we were going to use `date-fns`. I forgot what the api looks like but it probably won't be that hard.

aomi (approved):

PR #193, opened 2021-06-07
Collapse sections on calendar sidebar

### LINKED ISSUE #166: Add chevron button to saved course card
Add a chevron button to the saved course cards that appear on the righthand side of the timetable. Currently the only way to truncate a courses sections is to deselect it which completely removes it from the calendar. This is nice, but for courses like MATH 100 in the fall with a million sections it can get really crowded despite users wanting to see courses on the timetable still.

Adding a chevron button should: allow the user to minimize aa course on the sidebar so the card still appears, but the sections are hidden.
Sorry for a bad mock-up.

zthh626:
## Overview

Closes #166


FlyteWizard:
@keithradford I'm not sure if you folks want to create a new issue for this or try to solve it in this PR, but currently the solution here can't be utilized by keyboard users. I unfortunately do not know how I would get this to work for keyboard users. Maybe having the `Course Name` be clickable?

keithradford:
a couple nits about UX:

- can the title of the course be clickable along with the subject & code? Maybe wrap the VStack that both are a part of in a link?
- When hovering course & title I think it would be helpful if the text got underlined
- You can give stuff like text the prop `tabIndex={0}` which makes it accessible via keyboard
- right now it defaults to minimized, I think it should default to being open?
- lastly, can we flip the chevrons so it's pointing down when minimized and pointing up when open?

zthh626:
> a couple nits about UX:
> 
> * can the title of the course be clickable along with the subject & code? Maybe wrap the VStack that both are a part of in a link?
> * When hovering course & title I think it would be helpful if the text got underlined
> * You can give stuff like text the prop `tabIndex={0}` which makes it accessible via keyboard
> * right now it defaults to minimized, I think it should default to being open?
> * lastly, can we flip the chevrons so it's pointing down when minimized and pointing up when open?

`right now it defaults to minimized, I think it should default to being open?` for this point, it saves to local storage, so if it was previously minimized it will stay minimized when page is refreshed. Is it still doing this when adding a new course?

src/pages/scheduler/components/CourseCard.tsx
```diff
@@ -68,14 +88,29 @@ export function CourseCard({
...
+              spacing="0"
+              py="2"
+              tabIndex={0}
+              as={Link}
+              to={`/calendar/${term}/${subject}?pid=${pid}`}
+            >
+              <Text
+                fontSize="lg"
+                fontWeight="bold"
+                _hover={{
```
keithradford, line 103: If you add the `_hover` prop to the VStack will it make both the subject and title underline?

src/pages/scheduler/components/CourseCard.tsx
```diff
@@ -68,14 +88,29 @@ export function CourseCard({
...
-            <Text fontSize="sm" fontWeight="normal">
-              <Skeleton isLoaded={!loading}>{data?.title ?? ''}</Skeleton>
-            </Text>
-          </VStack>
+          <Flex grow={1}>
+            <VStack
+              alignItems="start"
+              spacing="0"
+              py="2"
+              tabIndex={0}
```
keithradford, line 96: I just tested this, it doesn't work when `tabIndex` is set on the VStack, but it does work if you set it on one of the `Text` children. Just setting it on one should be fine

src/lib/hooks/useSavedCourses.ts
```diff
@@ -173,6 +175,20 @@ export const useSavedCourses = (): SavedCourses => {
       data.map((course) => {
         if (equals(course, existingCourse)) {
           course.selected = selected;
+          if (!selected) course.showSections = false;
+          else course.showSections = true;
```
keithradford, line 179:
```suggestion
          course.showSections = selected;
```
does this work? If not it could also be
```suggestion
          course.showSections = selected ? true : false;
```

src/pages/scheduler/components/CourseCard.tsx
zthh626: Yep, I set it to the VStack.

src/lib/hooks/useSavedCourses.ts
zthh626: Changed to `course.showSections = selected ? true : false;`

src/pages/scheduler/components/CourseCard.tsx
zthh626: Moved tabIndex to just the course code and subject.

keithradford (approved):
LGTM with 1 nit, I think the chevron icons could be a _little_ bit bigger. Nothing crazy but if it could fill the button space a bit more than 👌

PR #195, opened 2021-06-07
Sid/git repo header

### LINKED ISSUE #163: Link to GitHub repo from header
Add a button (github icon button) on the far right-side of the header. Should link to the github repo!

siddharthpk:
## Overview

Closes #163  

Added a redirect to the Github repo using an icon button. 

Before 

courseup-before

After

courseup-after

### DIFF
--- a/src/app/header/Header.tsx
+++ b/src/app/header/Header.tsx
@@ -15,6 +15,7 @@ import { Link } from 'react-router-dom';
 
 import { useSessionStorage } from '../../shared/hooks/storage/useSessionStorage';
 
+import { ExternalLinks } from './components/ExternalLinks';
 import { NavButtons } from './components/NavButtons';
 import { Search } from './components/SearchBar';
 import { TermButtons } from './components/TermButtons';
@@ -65,6 +66,9 @@ export function Header({ onSearchChange }: HeaderProps): JSX.Element {
         </GridItem>
         <GridItem colStart={3}>
           <TermButtons />
+        </GridItem>
+        <GridItem colStart={4}>
+          <ExternalLinks />
         </GridItem>
       </Grid>
     </Box>
--- a/src/app/header/components/ExternalLinks.tsx
+++ b/src/app/header/components/ExternalLinks.tsx
@@ -1,0 +1,12 @@
+import { IconButton } from '@chakra-ui/react';
+import { useState } from 'react';
+import { DiGithubBadge } from 'react-icons/di';
+
+export function ExternalLinks(): JSX.Element | null {
+  const [isOpen, setIsOpen] = useState(false);
+  const open = () => {
+    setIsOpen(!isOpen);
+    window.open('https://github.com/VikeLabs/courseup');
+  };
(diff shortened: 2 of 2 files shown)


src/app/header/components/ExternalLinks.tsx
```diff
@@ -0,0 +1,12 @@
...
+import { useState } from 'react';
+import { DiGithubBadge } from 'react-icons/di';
+
+export function ExternalLinks(): JSX.Element | null {
+  const [isOpen, setIsOpen] = useState(false);
+  const open = () => {
+    setIsOpen(!isOpen);
+    window.open('https://github.com/VikeLabs/courseup');
+  };
+  return <IconButton aria-label="GitHub" icon={<DiGithubBadge />} borderRadius="50%" name="github" onClick={open} />;
```
keithradford, line 11: Instead of hooking into the state, you can add the props `as="a" href="https://github.com/VikeLabs/courseup"` which will do the same thing. Additionally, if you do it this way, you can add `target="_blank"` so clicking the link opens a new tab.

src/app/header/Header.tsx
```diff
@@ -66,6 +67,9 @@ export function Header({ onSearchChange }: HeaderProps): JSX.Element {
         <GridItem colStart={3}>
           <TermButtons />
         </GridItem>
+        <GridItem colStart={4}>
+          <ExternalLinks />
```
keithradford, line 71: If you opt for going with the props over the `useState` approach, this doesn't have to be it's own component

src/app/header/components/ExternalLinks.tsx
keithradford: nit: you can use the `isRound` prop instead of doing `borderRadius="50%"` ([chakra docs](https://chakra-ui.com/docs/form/icon-button#props) for reference)

keithradford: my fault for not creating a mock-up to base styling off of, but if we can aim to make it look like the attached screenshot that would be 🔥 
Screen Shot 2021-06-07 at 5 09 50 PM

src/app/header/Header.tsx
aomi: We _shouldn't_ consider this approach as there isn't a need for `useState` and complicates the code unnecessarily let alone accessibility issues I imagine.

src/app/header/components/ExternalLinks.tsx
aomi: The black colour scheme would be better for a11y.
https://contrastchecker.online/000000/82cbee whereas the white variant wouldn't pass.

siddharthpk: Thanks for the feedback guys! I'll change the code accordingly and request another review.

keithradford:
@siddharthpk Hey! Just checking in to see the status of this PR. Are you still working on this? (no rush, just a check-in)

siddharthpk:
Oh yeah, last and this week has been midterms so gonna complete this during the weekend.

aomi:
Closing due to inactivity. More than welcome to open a new PR if the issue is still open.

PR #206, opened 2021-06-10
feat: setup experimental jest-axe 

aomi:
Setups up testing in the frontend CI rules (for non main branches for now) and setup jest-axe.

Fixes the preview builds as well :crossed_fingers:


src/setupTests.ts
```diff
@@ -3,3 +3,6 @@
 // expect(element).toHaveTextContent(/react/i)
 // learn more: https://github.com/testing-library/jest-dom
 import '@testing-library/jest-dom';
+import { toHaveNoViolations } from 'jest-axe';
```
keithradford, line 6: is there a reason you're just adding the one `expect` method? Not that we need more right now but if we ever did in the future

keithradford (approved):
LGTM, just the one question

src/setupTests.ts
aomi: I think that's the only one from `jest-axe`

PR #207, opened 2021-06-10
feat: add git sha as static asset for frontend

aomi:
Adds a SHA as a static file to the application. This let's developers check which version of the application they are hitting.


keithradford (approved):
LGTM

aomi:
This doesn't work on production :face_with_head_bandage: oops....

PR #211, opened 2021-06-11
change overflow hidden to auto

### LINKED ISSUE #209: Bug: Overflow Hidden on Landing Page
**Describe the bug**

The landing page has overflow set to hidden, and the text is not scrollable, meaning text gets cut off. 

**To Reproduce**

1.  Visit https://courseup.vikelabs.ca.
2. View page in "non standard" format.
3. View that the text gets cut off. 

**Expected behaviour**

The text is not cutoff and is still viewable/scrollable.

**Screenshots**

Capture d’écran, le 2021-06-10 à 06 43 41

 - OS: macOS
 - Browser: Safari
 - Version: Latest

chsunwoo1002:
## Overview

Closes #209 

website cut off the text because overflow is set to hidden.
Change the setting to auto

Screen Shot 2021-06-11 at 9 13 04 AM

### DIFF
--- a/src/index.css
+++ b/src/index.css
@@ -1,3 +1,3 @@
 body {
-  overflow: hidden;
+  overflow: auto;
 }


keithradford (approved):
LGTM. 

The original reason we added `hidden` was to fix 2 scroll-bars showing up on the right-side (on the main content and for the whole page) so this fixed it. Verified and your changes seem to still work for that 👍 

Thanks! :)

aomi (changes requested):
This doesn't fix it. You get scrolling as a result. If you take out the `height` property somewhere near this stuff (I forgot where I'm on mobile right now)

keithradford:
> This doesn't fix it. You get scrolling as a result. If you take out the `height: 100%` property somewhere near this stuff (I forgot where I'm on mobile right now) it will fix this problem.

You get scrolling, but I didn't think that was that big of a deal. I guess we can aim to make it more responsive instead?

aomi:
Going to close due to inactivity.

PR #221, opened 2021-06-17
refactor: focus effects

keithradford:
## Overview

Made a series of a11y changes. Blue focus border now only appears when tabbing through page which maintains good a11y for keyboard users while increasing the aesthetic for mouse users. Also made the `CourseUp` text in the header tabable so  keyboard users can access and click it.
Relevant article about the :focus-visible trick I used [here](https://css-tricks.com/the-focus-visible-trick/).

Additionally, I added a `theme.tsx` file which can allow us to add custom theme and styling in the future.  (dark mode 😉 )


aomi (approved):
If @FlyteWizard is cool with these changes then LGTM.

keithradford:
Judging by the 👍 reaction it's a LGTM, gonna merge

PR #231, opened 2021-06-23
feat: Combine Staging/Production into single workflow

aomi:
This PR removes the staging related steps from the CI (frontend/backend) workflows and creates a new workflow called `deployment.yml`. 

This workflow builds and lints the frontend/backend in parallel as well as the prod/staging React builds in parallel. The staging version will be deployed directly but the production one will wait until a user reviews and accepts the deployment. 

Given these improvements, it should cut down on the deployment time and involve less manual intervention when deploying to production.

Future improvement is to add notifications to those involved in the deployment via Discord perhaps :eyes:


keithradford (approved):
LGTM

PR #232, opened 2021-06-24
fix: artifacts names and step names

aomi:
Fixes the artifact names in the deployment script as they were mismatched. Also makes the build step names more descriptive.


keithradford (approved):
🤞

PR #240, opened 2021-07-06
feat: Allow retrieving course details using term, subject and code

### LINKED ISSUE #100: Allow getting course details without using a pid
In order to use the `/courses/:term/:pid` (course details) endpoint right now, the `pid` is required. This causes an issue with the frontend where the `pid` + `term` cannot be used to retrieve section information. This results in the sections to be requested requires the details to be loaded in first which makes the frontend roughly twice as slow. We can fix this by using the `pid` stored in Firebase and allow fetching the course details with just the `term`, `subject` and `code` ie (` 202009 CSC 100`).

aomi:
## Overview
Closes #100 

Adds a new controller route to get course details using the format `{term}/{subject}/{code}`.
Note that this PR does not remove the old route for now but should be deprecated once the frontend is moved off of it.

This PR also modifies the local development for the backend to support using a local Firestore instance better.

To use the new local development Firestore:
1. `npm ci` in the `functions` directory. If you don't have the pre-requisuite database data ready it will download it when you start the dev server. 
2. Run `npm run db:start`. This will start the database. In another terminal run `npm run dev`. This will start the backend server.
3. To point the frontend to use the local backend, edit the `.env` file and change the `REACT_APP_BACKEND` from `staging` to `local`.


keithradford (approved):
LGTM!

keithradford:
One thing I noticed is subject is case-sensitve, i.e. `/courses/202109/csc/111` won't work.

keithradford (approved):
LGTM, working as expected 👌

PR #241, opened 2021-07-06
Bug fix: Subject cards now say (Not offered) if "Show in Session" filter button is used

### LINKED ISSUE #194: Subjects with no courses appear when using `Only Show Courses in Session`
**Describe the bug**

When using the `Only Show Courses in Session` filter, subjects with no courses in that session still appear in the subjects list.

**To Reproduce**

1. Go to https://courseup.vikelabs.ca/calendar/202105/
2. Click on the `Only Show Courses in Session` button
3. Click on ACAN, for instance

**Expected behavior**

With the filter enabled, subjects with no courses in the current session should not appear.

**Screenshots**

UIRumZgEva

 - OS: [e.g. iOS] Windows 10
 - Browser [e.g. chrome, safari] Microsoft Edge
 - Version [e.g. 22] Latest

**Additional context**

This may not be technically a bug but instead just the result of how the filter functions (since it's not an `Only Show Subjects in Session` filter), and this could instead be labeled as a possible QOL enhancement. If this is the case I can update the issue.

shergillmanav19:
made subject cards unclickable and added a warning hover for subjects that are not offered for that specific term

## Overview

Closes #194

### DIFF
--- a/src/common/sidebar/components/Card.tsx
+++ b/src/common/sidebar/components/Card.tsx
@@ -1,6 +1,6 @@
 import { PropsWithChildren, useCallback } from 'react';
 
-import { WarningTwoIcon, ChevronRightIcon, AddIcon, InfoOutlineIcon, CloseIcon } from '@chakra-ui/icons';
+import { ChevronRightIcon, AddIcon, InfoOutlineIcon, CloseIcon, NotAllowedIcon } from '@chakra-ui/icons';
 import { Box, Text, Flex, VStack, IconButton, Tooltip } from '@chakra-ui/react';
 import { Link, useParams } from 'react-router-dom';
 
@@ -60,7 +60,8 @@ export function Card({
   const { addCourse, deleteCourse, contains } = useSavedCourses();
 
   const courseIsSaved = pid && contains(pid, term);
-
+  const isSessionSubjectSet = inSessionSubject !== undefined;
+  // const isInSessionSubjectSet = inSessionSubject === undefined;
   const handleBookmarkClick = useCallback(() => {
     if (code && pid) {
       if (!courseIsSaved) {
@@ -82,15 +83,7 @@ export function Card({
         <Box>
           <IconButton
             aria-label="Select course"
-            icon={
-              inSessionSubject ? (
-                <ChevronRightIcon />
-              ) : (
-                <Tooltip label="No courses are offered in this term for this subject!" aria-label="A tooltip">
-                  <WarningTwoIcon color="red" />
-                </Tooltip>
-              )
-            }
+            icon={inSessionSubject ? <ChevronRightIcon /> : <NotAllowedIcon />}
             size="md"
             background="null"
             _hover={{ bg: 'none' }}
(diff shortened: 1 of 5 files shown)


aomi:
Not sure why the preview link isn't showing but here it is:
https://staging-clockwork--courseup-ba4f8o0z.web.app/scheduler/202105

src/common/sidebar/containers/SidebarContainer.tsx
```diff
@@ -52,9 +52,28 @@ export function SidebarContainer({ searchQuery, term }: SidebarContainerProps):
   const loading = subjectsLoading || coursesLoading;
 
   // sorts the list of subjects alphabetically
-  const sortedSubjects = useMemo(() => subjects?.sort((a, b) => (a.subject > b.subject ? 1 : -1)), [subjects]);
+  let sortedSubjects = useMemo(() => subjects?.sort((a, b) => (a.subject > b.subject ? 1 : -1)), [subjects]);
```
aomi, line 55: Maybe extend the type of `subjects[]` with a `inSession: boolean` instead?

src/lib/fetchers.tsx
```diff
@@ -112,6 +112,7 @@ export interface Seat {
 export interface KualiSubject {
   subject: string;
   title: string;
+  inSession?: boolean;
```
aomi, line 115: This file isn't stable (it's generated) so I suggest making a type that extends this one instead.

keithradford:
How about something like this design wise? It emulates how chakra deals with disabled buttons by basically just butting a grey tint over it. Additionally I think it should have this appearance permanently, not just on hover. Lastly, the pointer on hover should be `not-allowed` or just the default cursor so it doesn't appear clickable.

Screen Shot 2021-07-06 at 5 34 30 PM

src/common/sidebar/containers/SidebarContainer.tsx
```diff
@@ -48,12 +49,34 @@ export function SidebarContainer({ searchQuery, term }: SidebarContainerProps):
...
+        subject.inSession = false;
+        return subject;
+      }
+    });
+  } else {
+    sortedSubjects = sortedSubjects?.filter((subject) => {
+      subject.inSession = true;
+      return subject;
+    });
+  }
```
aomi, line 79: It would be better to add this `filter` logic within the `sortedSubjects` / `computeParsedCourses` as this avoids sidesteps the optimization of the `useMemo`.

src/lib/types.ts
```diff
@@ -0,0 +1,6 @@
+import { KualiSubject } from './fetchers';
+
+export interface KualiSubjectInSession {
+  inSession?: boolean;
+}
+export type InSessionSubject = KualiSubject & KualiSubjectInSession;
```
aomi, line 6: This can be combined into one type declaration. Also, I would suggest not calling it `InSessionSubject` as we might use it for other stuff :)

src/common/sidebar/containers/SidebarContainer.tsx
```diff
@@ -48,13 +49,29 @@ export function SidebarContainer({ searchQuery, term }: SidebarContainerProps):
     term: term as Term,
     queryParams: { in_session: filter },
   });
+  // re assign the type of subjects so inSession property is available
+  const inSessionSubjects: InSessionSubject[] | null = subjects;
```
aomi, line 53: This isn't required since you're mapping against the original thing. You can just return a `InSessionSubject` from the `map` below.

src/common/sidebar/components/Card.tsx
```diff
@@ -101,32 +120,70 @@ export function Card({ subject, title, code, selected, schedule, pid }: PropsWit
...
+            <Text fontSize="lg" fontWeight="bold" p={0} m={0}>
+              {subject} {code}
+            </Text>
+            <Text fontSize="sm" fontWeight="normal" p={0} m={0}>
+              {title}
+            </Text>
+          </VStack>
+          {buttons(code, schedule)}
+        </Flex>
+      </Box>
```
aomi, line 186: Can this be combined and just conditionally pass in the props?

aomi (commented):
Some visual issues I'm seeing:
image
The gray background colouring appears to be spilling over the defined card. This is due to the usage of the same gray so maybe switching it to be lighter/darker will help clear that up?

Also, I don't think the hover against the text needs to be present when a card is disabled.

The warning icon is low resolution. While I really like the tool-tip the icon usage seems a bit out of place. I'd maybe side on removing it for and add something like this once we've refined it a bit more. Interested in hearing your thoughts as well as @keithradford and @FlyteWizard

src/common/sidebar/components/SubjectsList.tsx
```diff
@@ -27,18 +27,28 @@ export function SubjectsList({ term, subjects }: SubjectsListProps): JSX.Element
...
+              <Card subject={subject.subject} inSessionSubject={subject.inSession} title={subject.title} />
+            </LinkBox>
+          );
+        } else {
+          return (
+            <Box key={index}>
+              <Card subject={subject.subject} inSessionSubject={subject.inSession} title={subject.title} />
+            </Box>
+          );
+        }
```
aomi, line 50: This can use tenary conditional so you can remove the `if`s and `returns`.

src/common/sidebar/components/Card.tsx
```diff
@@ -103,16 +122,36 @@ export function Card({ subject, title, code, selected, schedule, pid }: PropsWit
...
+            ? undefined
+            : inSessionSubject
+            ? undefined
+            : '#d3d3d3'
+          : inSessionSubject === undefined
+          ? 'white'
+          : inSessionSubject
+          ? 'white'
+          : '#d3d3d3'
+      }
```
aomi, line 137: This seems a bit extreme and could likely be simplified somehow. Or even just moving this into separate variables so it's easier to read would be great.

aomi:
Given the outdated nature and lack of user reports about this issue closing.

PR #247, opened 2021-07-16
feat(lint): make unused-import an error

keithradford:
# Description

Makes the `unused-import` linting warning an error, as it was one of the most common linting warnings we'd se in PRs. Should make it a lot more obvious when the error is present. Additionally, moved the eslint rules into a separate config file to decrease a bit of clutter in `package.json` which is getting a bit long, especially with the Jest changes in #242 

Closes #XX


aomi:
@keithradford could you make a lint error so we can verify that it's working as intended?

keithradford:
@aomi 🎉

aomi (approved):
I'll let you get your deploy fix in I guess.

PR #252, opened 2021-07-25
feat: Timetable sharing backend

### LINKED ISSUE #250: Create backend for timetable sharing

keithradford:
# Description

Closes #250 

Created the backend for the timetable sharing functionality. Timetables will be stored in our database with the following structure:
```ts
{
  term: string;
  slug: string;
  courses: {
    subject: string;
    code: string;
    pid: string;
    lecture?: string;
    lab?: string;
    tutorial?: string;
    color: string;
  }[]
}
```
which emulates how courses are stored in local storage. The only difference here is term has been moved out of the course attribute and into the upper scope since all courses will be in the same term, and the `selected` boolean is not relevant for the backend thus is not included here. 

This feature adds two endpoints, one for fetching a timetable and one for creating a timetable. 

- GET `/api/timetables/:slug`
   -  Fetches a timetable based on a given slug, returns an object with the structure above minus the slug.
- POST `/api/timetables/`
   -  Creates a timetable to be stored in the data base.
   - Requires a `term` and `courses` as part of the payload, as defined above.
   - Generates a slug by *sorting and hashing* the courses array and term as an object.
   - Validates data as follows:
     -  `term` must be the `Term` type, which ensures it's of the form `202109`, `202001`, etc.
     - `courses` must be <= 12 to avoid storing unnecessarily large data
     - each lecture, lab, and tutorial attribute must match a RegEx pattern (of the form `A01`, `T02`, `B11`, etc.)
     - The `color` attribute must be a valid hex format
     - The course must be in the database
     - The `pid` attribute from the payload must match the `pid` of the course it finds in the database.

## Checklist

- [x] The code is documented.

## General Comments

~TODO: Add tests, potentially more data validation.~


functions/package.json
```diff
@@ -33,9 +33,11 @@
     "@firebase/testing": "^0.20.11",
     "@types/async": "^3.2.6",
     "@types/body-parser": "^1.19.0",
+    "@types/crypto-js": "^4.0.2",
```
aomi, line 36: Don't think this is needed.

functions/src/db/collections.ts
```diff
@@ -26,6 +28,13 @@ export type SectionDoc = {
   instructors?: string[];
 };
 
+export type TimetableDoc = {
+  term: Term;
+  courses: TimetableCourse[];
+  createdAt: Date;
+  updatedAt?: Date;
+};
```
aomi, line 36: add a `hash: string` value so we can query via it.

functions/src/index.ts
```diff
@@ -49,6 +51,28 @@ app.get('/openapi.json', async (req, res) => {
...
+  }
+  if (err instanceof Error) {
+    return res.status(500).json({
+      message: 'Internal Server Error',
+    });
+  }
+
+  next();
+});
+
```
aomi, line 75: Don't think we want to expose this on production.

functions/src/timetables/Timetable.controller.ts
```diff
@@ -0,0 +1,56 @@
...
+@Route('timetables')
+export class TimetablesController extends Controller {
+  @Response(404, 'Not found')
+  @Get('{slug}')
+  public async getTimetable(@Path() slug: string): Promise<Timetable | void> {
+    // set the Cache-Control for 24h.
+    this.setHeader(
+      'Cache-Control',
+      `public, max-age=${3600}, s-max-age=${3600}, stale-while-revalidate=${30}, stale-if-error=${60}`
+    );
```
aomi, line 33: This can be cached longer as generally we don't change the timetable data.

functions/src/timetables/Timetable.model.ts
```diff
@@ -0,0 +1,48 @@
...
+  code: string;
+  /**
+   * The PID of the course.
+   * @example "ByS23Pp7E"
+   */
+  pid: string;
+  /**
+   * The selected lecture section of the course.
+   * @example "A02"
+   * @pattern ^A\d{2}$
```
aomi, line 22: this is pretty cool

functions/src/timetables/Timetable.service.ts
```diff
@@ -0,0 +1,68 @@
...
+    )
+      return false;
+
+    // validate course is in database
+    const doc = await get(
+      CoursesCollection,
+      constructSectionKey(term, course.subject, course.code)
+    );
+    if (!doc || doc.data.pid !== course.pid) return false;
+  }
```
aomi, line 65: Using `Promise.all` would allow validation them all at once rather than one at a time.

functions/src/timetables/Timetable.controller.ts
keithradford: what do you think would make sense? I could do a week or something

keithradford:
@aomi Note [this test ](https://github.com/VikeLabs/courseup/pull/252/files#diff-f53626f975e30d655eec894cde06fb5700a8faa7cafb80d6aae5138e97f22132R100 )is failing but I honestly don't know why. The hash query is matching for everything in the mock DB so it's not being added despite a unique hash

functions/src/timetables/Timetable.controller.ts
aomi: Keep it this for now I think since it's the first release :shrug:

functions/src/timetables/Timetable.controller.ts
```diff
@@ -0,0 +1,56 @@
...
+    if (!data) {
+      this.setStatus(404);
+      return;
+    }
+    return data;
+  }
+
+  @Response<ValidateErrorJSON>(422, 'Invalid data')
+  @SuccessResponse('201', 'Created')
+  @Post()
```
aomi, line 44: I think a `PUT` would be more appropriate given it's idempotent.

functions/src/timetables/Timetable.service.ts
```diff
@@ -0,0 +1,90 @@
...
+      )
+        return false;
+
+      // validate course is in database
+      const doc = await get(
+        CoursesCollection,
+        constructSectionKey(term, course.subject, course.code)
+      );
+
+      if (!doc || doc.data.pid !== course.pid) return false;
```
aomi, line 82: I'm not sure if the `pid` check is needed here?

functions/src/timetables/Timetable.service.ts
```diff
@@ -0,0 +1,90 @@
...
+
+export async function hasValidCourses(
+  courses: TimetableCourse[],
+  term: Term
+): Promise<boolean> {
+  // 12 is the max length of courses we will allow for a timetable in the database
+  if (courses.length > 12 || courses.length === 0) return false;
+
+  const validations = await Promise.all([
+    ...courses.map(async (course) => {
```
aomi, line 68: Why not just do `Promise.all(courses.map)`?

functions/src/timetables/__tests__/Timetable.controller.test.ts
```diff
@@ -0,0 +1,86 @@
...
+  slug: 'abc123',
+  term: '202109',
+  courses: [
+    {
+      subject: 'ACAN',
+      code: '225',
+      pid: 'ByS23Pp7E',
+      lecture: 'A02',
+      lab: 'B02',
+      color: '#48BB78',
```
aomi, line 29: Does validation run on these attributes? It would be good to test if weird values cannot be set in these.

functions/src/timetables/Timetable.service.ts
keithradford: it's essentially just an extra layer of data validation

functions/src/timetables/__tests__/Timetable.controller.test.ts
keithradford: it's mocked so don't think so

functions/src/timetables/Timetable.service.ts
aomi: Fair enough

functions/src/timetables/__tests__/Timetable.controller.test.ts
```diff
@@ -0,0 +1,86 @@
...
+      it('should return a 400 status', async () => {
+        const response = await post('/timetables').send(timetable);
+        expect(response.statusCode).toBe(400);
+      });
+
+      it('should return void', async () => {
+        const response = await post('/timetables');
+        expect(response.body).toStrictEqual({});
+      });
+    });
```
aomi: Can you add a failure case when the timetable contains invalidly formed data (which would fail the pattern on the model)?

aomi (commented):
I tried deploying this on a demo env but it won't build properly :|

functions/src/timetables/Timetable.service.ts
```diff
@@ -0,0 +1,99 @@
...
+    courses.map(async (course) => {
+      if (
+        (course.lab &&
+          (course.lab.length > 1 || !labRegex.test(course.lab?.[0]))) ||
+        (course.lecture &&
+          (course.lecture.length > 1 ||
+            !lectureRegex.test(course.lecture?.[0]))) ||
+        (course.tutorial &&
+          (course.tutorial.length > 1 ||
+            !tutorialRegex.test(course.tutorial?.[0])))
```
aomi, line 81: nit: I would use optional chaining syntax here so you can clean up the checks.

functions/src/timetables/__tests__/Timetable.controller.test.ts
```diff
@@ -0,0 +1,180 @@
...
+  }
+  if (err instanceof Error) {
+    return res.status(500).json({
+      message: 'Internal Server Error',
+    });
+  }
+
+  next();
+});
```
aomi, line 48: nit: can we move this into the middlewares directory since it's used in a bunch of spots.

aomi (approved):
small nit but :crossed_fingers:

functions/src/timetables/Timetable.service.ts
keithradford: not sure if this really cleans it up. optional chaining makes the type `<T> | undefined` and we can't compare `undefined` to a number for the > test. Also, can't pass `undefined` into `RegExp.test()`

PR #253, opened 2021-07-26
feat(experiment): add dark mode

keithradford:
# Description

Adds dark mode 🌚 ! Chakra offers a pretty good built-in dark mode, so I've modified our app to take advantage of that as much as possible. What this means is:

- Any "hard coded" colours have been removed (i.e. places the `color` or `bgColor` prop were used)
  - Chakra will default text to black in light mode, white in dark mode as well as backgrounds to white and black in each respective mode which means these don't need to be explicitly stated anywhere as props, which they were previously.
- I have opted to use the `colorScheme` prop wherever possible, which just about all Chakra components have
  - The `colorScheme` automatically changes the text and prop colour based on what mode the user is in which is great
-  Any areas of the code that don't have automatic chakra support, such as custom colours (we use gray in lots of areas) and the timetable have custom styles in `theme.tsx`
  - `theme.tsx` allows for easy custom "dark mode" styles, so any edge cases were added here.
- There is also a hook which allows us to conditionally check if the page is in dark mode which is used in some situations.   

## Screenshots

There's a lot, so not gonna bother 😅 

## Checklist

- [ ] The code is documented.

## General Comments

Please take a look at the preview build and make sure I didn't miss anything or mess any styling up (light and dark mode). Furthermore, I would appreciate if you notice there's poor contrast in certain areas!


src/pages/home/containers/Landing.tsx
```diff
@@ -33,10 +36,10 @@ export function Landing() {
           </Heading>
           <Text fontSize="1.6em">CourseUp makes it simple to browse and schedule UVic Courses</Text>
           <Flex alignItems="center" mt="1rem" as="a" w="fit-content" target="_blank" href="https://www.vikelabs.ca">
-            <Text color="#4C6EA5" fontSize={{ base: '1em', md: '1.5em' }}>
+            <Text fontSize={{ base: '1em', md: '1.5em' }} color={isDarkMode ? '#799EC1' : '#4C6EA5'}>
               <span>{hands[handIndex]}</span> Built by students @
             </Text>
-            <Text fontWeight="bolder" fontSize="1.75em" color="#222B49" ml="1">
+            <Text fontWeight="bolder" fontSize="1.75em" color={isDarkMode ? '#B0C3DA' : '#222B49'} ml="1">
```
aomi, line 42: A thought I had is it might be good to define these colours in the theme instead of throwing hex values around :thinking:

src/pages/scheduler/components/SchedulerCalendar.tsx
```diff
@@ -283,7 +285,7 @@ export function SchedulerCalendar({ calendarEvents }: SchedulerCalendarProps): J
       views={['work_week']}
       date={selectedDate}
       eventPropGetter={eventStyleGetter}
-      slotPropGetter={slotPropGetter}
+      slotPropGetter={(date) => slotPropGetter(date, isDarkMode)}
```
aomi, line 288: Might be good to make a `handleSlotPropGetter` and inject the `isDarkMode` within that.

src/lib/hooks/useIsDarkMode.ts
```diff
@@ -0,0 +1,7 @@
+import { useColorMode } from '@chakra-ui/react';
+
+export const useIsDarkMode = (): boolean => {
+  const { colorMode } = useColorMode();
+
+  return colorMode !== 'light';
+};
```
aomi, line 7: nit++: Maybe maybe a hook called `useColorMode` which returns a `mode` function with the same functionality as the `mode` in theme tools? Then you can avoid a lot of the logic (basic but repeated many times) and keep consistent with theme `mode` usage.

keithradford: I will definitely document all this stuff once it goes through

aomi (approved):

PR #256, opened 2021-08-04
fix: sections being hidden for courses added before this feature

keithradford:
# Description

#193 added the option to collapse courses on the timetable sidebar, however the edge case where users who had timetables made prior to this additional attribute being added was not addressed. Currently on staging if you already had courses added, they will default to be minimized since `showSections` is undefined, which isn't the end of the world but I'd rather we avoid this.


zthh626 (changes requested):
Tested by pasting local storage without showSections and they all default to being shown, but the chevron doesn't do anything and is the opposite icon.

Screen Shot 2021-08-04 at 9 04 48 AM

where it should be:

Screen Shot 2021-08-04 at 9 05 52 AM

zthh626 (approved):

Issue #262, opened 2021-08-09
Clean-up z-index use


szeckirjr:
**Describe the Bug/Issue**

Currently, the header/banner z-index is set to 1000. This messes up with certain Chakra components (like modals, as they will be partly covered).

Chakra [recommends not exceeding z-index values of more than 1 or 2](https://chakra-ui.com/guides/z-index).

**Screenshots**
![image](https://user-images.githubusercontent.com/50898635/128651156-6509ceed-60a2-4d9f-89c3-fd879678ae21.png)

Schorcher5:
I think this issue has already been resolved. I attempted to recreate this issue with now success. The z index of the header/banner looks to be set to auto, not 1000

keithradford:
@Schorcher5 This probably got cleaned up at some point in another PR. Thanks for pointing it out!

PR #263, opened 2021-08-09
Timetable post flow UI

### LINKED ISSUE #258: Create shareable timetable POST flow on frontend
Implement the POST flow the the timetable sharing feature on the frontend. 
UI and flow can be found on [figma](https://www.figma.com/file/Lvep4lP1867IZUiVQHUtaz/CourseUp?node-id=515%3A10672)

### LINKED ISSUE #260: Add social media integration to timetable sharing
See mock-ups in the POST flow for where social media integrations will be on [figma](https://www.figma.com/file/Lvep4lP1867IZUiVQHUtaz/CourseUp?node-id=515%3A10672)

For this ticket, we will need to scope out which media platforms we want to use and explore their APIs to see if it is doable.
Feel free to add to this list:

- Facebook (messenger)
- WhatsApp
- Telegram
- Discord

szeckirjr:
# Description

This PR adds a "Share" button to the timetable view.
When clicked, it opens a modal that includes a review of all courses selected in the current term, and options to share the generate link using social media or copying.

Closes #258, Closes #260 

## Screenshots
image

## Checklist

- [ ] The code is documented.

## General Comments

The modal is shows under the banner/header when on mobile. This will require completing #262


src/pages/scheduler/components/ShareTimetableButton.tsx
```diff
@@ -65,21 +65,27 @@ const ShareCourseCard = (props: { course: SavedCourse }) => {
 const SelectedCoursesTable = (props: { term: string }) => {
   const { courses } = useSavedCourses();
 
+  const inSession_savedCourses = courses
+    .filter((course) => course.term === props.term)
+    .filter((course) => course.lecture != null || course.lab != null || course.tutorial != null);
```
aomi, line 70: Right idea but you can just `or` everything instead.

szeckirjr: you mean by taking out `!=null`? I thought it [wouldn't work](https://stackoverflow.com/questions/20940545/how-can-i-check-whether-an-optional-parameter-was-provided) for both `null` and `undefined` values if I didn't include that?

package.json
```diff
@@ -29,6 +29,7 @@
     "react-router": "^6.0.0-beta.0",
```
keithradford, line 29: adding comment here, but do you know why ther eis 36,000 additions to `package-lock.json` 🤔

src/pages/scheduler/components/ShareTimetableButton.tsx
```diff
@@ -0,0 +1,185 @@
...
+  TwitterIcon,
+  TwitterShareButton,
+  WhatsappIcon,
+  WhatsappShareButton,
+} from 'react-share';
+
+import { SavedCourse, useSavedCourses } from 'lib/hooks/useSavedCourses';
+import { getReadableTerm } from 'lib/utils';
+
+const InformationText = (props: { isSmallScreen: boolean; term: string }) => {
```
keithradford, line 40: I'd say this probably doesn't need to be it's own component

src/pages/scheduler/components/ShareTimetableButton.tsx
```diff
@@ -0,0 +1,185 @@
...
+  return (
+    <Alert status="info" borderRadius="10px">
+      {props.isSmallScreen ? <AlertIcon /> : null}
+      We've generated a link that you can share to allow people to view, compare, and import the courses and sections
+      you currently have selected for the {getReadableTerm(props.term)} term.
+    </Alert>
+  );
+};
+
+const ShareCourseCard = (props: { course: SavedCourse }) => {
```
keithradford, line 50: The way we structure props in custom components is usually like such
```suggestion
type Props = {
   course: SavedCourse;
}
const ShareCourseCard = ({ course }: Props) => {
```
this is called object de-structuring, which is generally nicer because we don't have to call `props.` everytime, and can reference the variables directly

src/pages/scheduler/components/ShareTimetableButton.tsx
```diff
@@ -0,0 +1,185 @@
...
+  return (
+    <Flex height="100%" direction="column">
+      <HStack w="100%" bg="#EDF2F7" justifyContent="center" p="0.2em" borderTopRadius="4px" padding="4px">
```
keithradford, line 53: See https://github.com/VikeLabs/courseup/wiki/Theme-Guide for theme guidelines, we should try and avoid hard coding in hexes for the sake of dark mode

keithradford: `p` and `padding` are the same props, I'd recommend just keeping `p` which is what our other components usually go with

src/pages/scheduler/components/ShareTimetableButton.tsx
```diff
@@ -0,0 +1,185 @@
...
+      </HStack>
+      <Button
+        size={props.isSmallScreen ? 'sm' : 'md'}
+        colorScheme="blue"
+        leftIcon={<CopyIcon />}
+        onClick={() => {
+          var el = document.getElementById('timetable_slug') as HTMLInputElement;
+          el.focus();
+          el.select();
+          document.execCommand('copy');
```
keithradford, line 118: https://chakra-ui.com/docs/hooks/use-clipboard is probably a nicer approach. You can see it used here https://github.com/VikeLabs/courseup/blob/main/src/pages/registration/components/RegistrationSection.tsx

src/pages/scheduler/components/ShareTimetableButton.tsx
```diff
@@ -0,0 +1,185 @@
...
+    <HStack justify="space-between" padding="5px" borderWidth="1px" borderColor="gray.300">
+      <HStack justify="center" flexGrow={4}>
+        {props.isSmallScreen ? <Icon boxSize="1.25em" as={HiLink} /> : undefined}
+        <Input id="timetable_slug" value={props.slug} variant="filled" isReadOnly={true} />
+      </HStack>
+      <Button
+        size={props.isSmallScreen ? 'sm' : 'md'}
+        colorScheme="blue"
+        leftIcon={<CopyIcon />}
+        onClick={() => {
```
keithradford, line 114: Typically we don't want to use arrow functions here and would probably define a function above called `handleClick` but since there is still API integration to go into slug generation, we can refactor this later

src/pages/scheduler/components/ShareTimetableButton.tsx
```diff
@@ -0,0 +1,185 @@
...
+      <WhatsappShareButton children={<WhatsappIcon size={50} round={true} />} url={props.slug} />
+      <TwitterShareButton children={<TwitterIcon size={50} round={true} />} url={props.slug} />
+    </HStack>
+  );
+};
+
+const CopyLinkUrl = (props: { isSmallScreen: boolean; slug: string }) => {
+  const toast = useToast();
+  return (
+    <HStack justify="space-between" padding="5px" borderWidth="1px" borderColor="gray.300">
```
keithradford, line 105: if `gray.300` looks fine on dark and light mode, that's ok but if not you can use the `useDarkMode` hook to switch colours

src/pages/scheduler/components/ShareTimetableButton.tsx
```diff
@@ -0,0 +1,185 @@
...
+      <HStack justify="center" flexGrow={4}>
+        {props.isSmallScreen ? <Icon boxSize="1.25em" as={HiLink} /> : undefined}
+        <Input id="timetable_slug" value={props.slug} variant="filled" isReadOnly={true} />
```
keithradford, line 108: i think you could just say `isReadOnly` without passing true (maybe not, but if you can it's a little more clean)

src/pages/scheduler/components/ShareTimetableButton.tsx
```diff
@@ -0,0 +1,185 @@
...
+        size={props.isSmallScreen ? 'sm' : 'md'}
+        colorScheme="blue"
+        leftIcon={<CopyIcon />}
+        onClick={() => {
+          var el = document.getElementById('timetable_slug') as HTMLInputElement;
+          el.focus();
+          el.select();
+          document.execCommand('copy');
+          toast({
+            title: 'Copied.',
```
keithradford, line 120: Would be good to say `Copied ___ to clipboard` to give users feedback that they copied the expected link maybe?

src/pages/scheduler/components/ShareTimetableButton.tsx
```diff
@@ -0,0 +1,185 @@
...
+    .filter((course) => course.term === term)
+    .filter((course) => course.lecture || course.lab || course.tutorial);
+
+  return (
+    <>
+      <Button
+        isDisabled={inSession_savedCourses.length > 0 ? false : true}
+        size="sm"
+        bg="blue.400"
+        color="white"
```
keithradford, line 166:
```suggestion
       colorScheme="blue"
```
will handle dark mode

keithradford (commented):
A couple comments on how I would structure this:

- I would turn this component into `ShareTimetableModal` as appose to `ShareTimetableButton` and then the component is just what you have from line 173-182, and the other defined comonents. Then you can just add the button and `isOpen` logic to the main scheduler content
- Within the `...../scheduler/components` directory I'd make a `/share` directory where we can store all components for this
- I would break this file down into more components. It's a bit long, and it's generally good practice to separate all your custom components into their own files. Keeps things clean. Of course if it's a super minor component, it's *probably* fine to leave, but use your judgement there :)

package.json
szeckirjr: hmm I honestly have no clue 😕 tried searching a bit but I really don't understand it enough to know how it could have 36k+ additions?

szeckirjr:
@keithradford thanks again for all the thorough feedback! I have pushed changes that I believe fixes all the problems addressed.

Let me know if you (or anyone) has any other things they suggest change! :)

src/pages/scheduler/components/share/SelectedCoursesCardList.tsx
```diff
@@ -0,0 +1,56 @@
+import { Box, Flex, Heading, HStack, VStack, Wrap, WrapItem, Text } from '@chakra-ui/react';
+
+import { useDarkMode } from 'lib/hooks/useDarkMode';
+import { SavedCourse } from 'lib/hooks/useSavedCourses';
+import { getReadableTerm } from 'lib/utils';
+
+type ShareCourseCard_Props = {
```
keithradford, line 7: nit: this could be `ShareCourseCardProps`, to avoid mixing camel case and underscores

src/pages/scheduler/components/share/SelectedCoursesCardList.tsx
```diff
@@ -0,0 +1,56 @@
...
+};
+
+const ShareCourseCard = ({ course }: ShareCourseCard_Props) => {
+  const mode = useDarkMode();
+
+  return (
+    <Flex height="100%" direction="column">
+      <HStack w="100%" bg={mode('gray.100', 'gray.600')} justifyContent="center" p="0.3em" borderTopRadius="4px">
+        <Heading size="xs">
+          {course.subject} {course.code}
```
keithradford, line 18: If you are never explicitly using `course` and only it's attributes, it's probably better to define `subject, code, color, lecture, lab, and tutorial` as separate props OR you can do object de-structuring on course

src/pages/scheduler/components/share/ShareLinkOptions.tsx
```diff
@@ -0,0 +1,78 @@
...
+      <EmailShareButton children={<EmailIcon size={50} round={true} />} url={slug} />
+      <FacebookShareButton children={<FacebookIcon size={50} round={true} />} url={slug} />
+      <TelegramShareButton children={<TelegramIcon size={50} round={true} />} url={slug} />
+      <WhatsappShareButton children={<WhatsappIcon size={50} round={true} />} url={slug} />
+      <TwitterShareButton children={<TwitterIcon size={50} round={true} />} url={slug} />
+    </HStack>
+  );
+};
+
+type CopyLinkUrl = {
```
keithradford, line 37:
```suggestion
type CopyLinkUrlProps = {
```
`CopyLinkUrl` is already defined below

src/pages/scheduler/components/share/ShareLinkOptions.tsx
```diff
@@ -0,0 +1,78 @@
...
+import { useDarkMode } from 'lib/hooks/useDarkMode';
+
+type SocialMediaButtons_Props = {
+  slug: string;
+};
+
+const SocialMediaButtons = ({ slug }: SocialMediaButtons_Props) => {
+  return (
+    <HStack justify="center">
+      <EmailShareButton children={<EmailIcon size={50} round={true} />} url={slug} />
```
keithradford, line 28:
```suggestion
      <EmailShareButton children={<EmailIcon size={50} round />} url={slug} />
```
small nit, not even sure if this will work

keithradford:
closing since we have new PRs

Issue #55, opened 2021-03-13
Improve and expand on section badges 


keithradford:
![firefox_Q7hQTLYdLQ.png](https://images.zenhubusercontent.com/5f973f7c87cad5309bf82827/5f1ab3f3-8564-4b1b-bc9b-d2f8a1f5df49)

Currently course sections display badges based on a few criteria such as `online`, `asynchronous`, and `SENG/CSC only` (see screenshot). However, the code is somewhat hacky and could be improved on by being made more generic and reliable, as well as hit more cases than the current ones.

For this ticket, you could take a look at the current implementation and think of ways it can be improved on and look through subjects other than SENG that have generic attributes which could be represented as a badge. Badges are key information we want to show users front and center, so consider this when looking. 
An example of why we should refactor the code would be that there is not a `synchronous` badge because using the current logic, `synchronous` would be found when `asynchronous` is present and you can see how that's wrong.

JoshoTheMosho:
I'd like to take on this issue as an ongoing thing to work on and better familiarize myself with everything 

# First change
Added on a 'Synchronous' tag to courses that were online but not asynchronous, kept it off of face-to-face courses to avoid clutter

# Before
![image](https://user-images.githubusercontent.com/48394084/125596705-e1104274-becd-4faf-afe9-875a75644858.png)

# After
![image](https://user-images.githubusercontent.com/48394084/125596835-6c17d644-65fa-4a29-9b61-830640cbe99e.png)

JoshoTheMosho:
ECE 216 and ECE 220 can be mapped out

PR #265, opened 2021-08-09
Josho/expand section badges

JoshoTheMosho:
# Description

Hard coded some additional badge info for courses
Adjusted colours to match information being conveyed (ex. Red for restrictions, blue for sync/async/blended etc.)

Addresses #55

## Screenshots

image
image
image

## General Comments

Looking to make this more dynamic with regex, however, UVic doesn't follow a simple template for restrictions and section comments


aomi:
@JoshoTheMosho you are using `==` instead of `===` #just-javascript-things so make sure you fix that.
Also, if you can list some courses for which these new badges will display that will make review a lot easier :+1:

src/common/header/components/GitHubButton.tsx
```diff
@@ -0,0 +1,21 @@
+import { IconButton } from '@chakra-ui/button';
+import { AiFillGithub } from 'react-icons/ai';
+
+export function GitHubButton(): JSX.Element {
+  return (
+    <IconButton
+      as="a"
```
keithradford, line 7: you can delete this file, it got added and then removed but git thinks you're adding it.

src/pages/calendar/components/Section.tsx
```diff
@@ -59,9 +59,15 @@ export function SectionInfo({
   seat,
   meetingTimes,
 }: SectionInfoProps): JSX.Element {
-  const isASYNC = additionalNotes?.indexOf('asynchronous') !== -1;
+  const isASYNC = additionalNotes?.indexOf('This course will be offered fully online and asynchronous') !== -1;
+  const isBLENDED = additionalNotes?.indexOf('a mix of “real-time” and asynchronous sessions') !== -1;
   const isSENG = additionalNotes?.indexOf('Reserved for BSENG students') !== -1;
   const isCSC = additionalNotes?.indexOf('Reserved for students in a Computer Science program') !== -1;
+  const isENGR = additionalNotes?.indexOf('Restricted to BEng students only') !== -1;
+  const isENGR_CSC = additionalNotes?.indexOf('Restricted to students in the Faculty of Engineering') !== -1;
```
keithradford, line 67: to avoid mixing camel case and underscores, maybe a good name for this stuff is `isENGRorCSC`, up to you though

src/pages/calendar/components/Section.tsx
```diff
@@ -82,16 +88,51 @@ export function SectionInfo({
                 Asynchronous
               </Badge>
             )}
+            {!isASYNC && !isBLENDED && instructionalMethod == 'online' && (
+              <Badge colorScheme="cyan" mx="1">
+                Synchronous
+              </Badge>
+            )}
+            {isBLENDED && (
```
keithradford, line 96: I'd say we could probably make these custom components that take in a boolean (like `isBLENDED`, `isASYNC`, etc.) and then conditionally render from that. Since this code is repeated so much it would clean it up.

aomi: We can likely move this to the the backend entirely but I think for now it's fine for now.

JoshoTheMosho:
Resolved most comments + meeting with Keith about conditional rendering

Courses + changes examples
image
image

image
image

src/pages/calendar/components/Section.tsx
```diff
@@ -59,9 +78,15 @@ export function SectionInfo({
...
-  const isASYNC = additionalNotes?.indexOf('asynchronous') !== -1;
+  const isASYNC = additionalNotes?.indexOf('This course will be offered fully online and asynchronous') !== -1;
+  const isBLENDED = additionalNotes?.indexOf('a mix of “real-time” and asynchronous sessions') !== -1;
   const isSENG = additionalNotes?.indexOf('Reserved for BSENG students') !== -1;
   const isCSC = additionalNotes?.indexOf('Reserved for students in a Computer Science program') !== -1;
+  const isENGR = additionalNotes?.indexOf('Restricted to BEng students only') !== -1;
+  const isENGRorCSC = additionalNotes?.indexOf('Restricted to students in the Faculty of Engineering') !== -1;
+  const isSCI = additionalNotes?.indexOf('Restricted to SCIENCE students') !== -1;
+  const isBENGorBSENG = additionalNotes?.indexOf('Restricted to BEng and BSEng students') !== -1;
+  const isENGRmulti = additionalNotes?.indexOf('Reserved for BME, BSEN, CENG, ELEC students') !== -1;
```
aomi, line 89: Since this is getting a bit long maybe we can consider refactoring to something like the following:
```
let rules be an array of rules where a rule might look like "a mix of real-time"
let badges be an array of badges where a badge object contains a name and colour atrribute 
for each rule in rules
  if found add name and colour to badges array
end
```
then using a `badges` array, we could render them out like:
```jsx
{badges.map(badge => <CourseBadge name={name} color={colour}/>)
```

JoshoTheMosho:
image

_Note:_ Want to look into course restrictions that aren't included in 'additional notes', possibly display those restrictions within those notes?
Also should that be left for the back end stuff since the restrictions aren't being fetched yet afaik?

src/pages/calendar/components/Section.tsx
```diff
@@ -59,9 +78,56 @@ export function SectionInfo({
...
+  ];
+
+  for (let i = 0; i < rules.length; i++) {
+    if (additionalNotes?.indexOf(rules[i]) !== -1) {
+      badges.push({
+        name: names[i],
+        color: 'red',
+      });
+    }
+  }
```
aomi, line 130: Could the `rules` and `names` be combined?
```js
const rules = [
  {rule: "Reserved for BSENG students", title: "SENG Only", color: "orange"}
]
const badges = rules.filter(r => additionalNotes?.indexOf(r) !== -1).map(r => {name: r.title, color: r.color})
```
Also, the rules/names should be defined outside the function. It doesn't make sense to declare them each time

src/pages/calendar/components/Section.tsx
```diff
@@ -77,21 +143,11 @@ export function SectionInfo({
...
-            {isCSC && (
-              <Badge colorScheme="yellow" mx="1">
-                CSC ONLY
-              </Badge>
-            )}
+            {badges.map((b) => (
+              <CourseBadge name={b.name} color={b.color}>
+                {b.name}
+              </CourseBadge>
+            ))}
```
aomi, line 150: This is missing a `key` prop. 
https://reactjs.org/docs/lists-and-keys.html

src/pages/calendar/components/Section.tsx
```diff
@@ -16,6 +18,33 @@ import { useDarkMode } from 'lib/hooks/useDarkMode';
...
+
+const ruleNames: string[][] = [
+  ['Reserved for BSENG students', 'SENG Only'],
+  ['Computer Science program.', 'CSC Only'],
+  ['BEng students', 'ENGR Only'],
+  ['Faculty of Engineering', 'ENGR/CSC Only'],
+  ['SCIENCE students', 'SCI Only'],
+  ['BEng and BSEng students', 'BEng/BSeng Only'],
+  ['BME, BSEN, CENG, ELEC students', 'BME/SENG/CENG/ELEC Only'],
+];
```
aomi, line 34: We should likely use objects here (typed) rather than an array here given it has some room for error.

src/pages/calendar/components/Section.tsx
```diff
@@ -59,9 +88,36 @@ export function SectionInfo({
...
+  }
+
+  for (let i = 0; i < ruleNames.length; i++) {
+    if (additionalNotes?.indexOf(ruleNames[i][0]) !== -1) {
+      badges.push({
+        name: ruleNames[i][1],
+        color: 'red',
+      });
+    }
+  }
```
aomi, line 120: Ideally this is more concise. If you can provide details why its "funky" it would help in the future.

src/pages/calendar/components/Section.tsx
```diff
@@ -77,21 +133,11 @@ export function SectionInfo({
...
-              <Badge colorScheme="orange" mx="1">
-                SENG ONLY
-              </Badge>
-            )}
-            {isCSC && (
-              <Badge colorScheme="yellow" mx="1">
-                CSC ONLY
-              </Badge>
-            )}
+            {badges.map((badges, index) => (
```
aomi, line 136: I'd use name here instead of index as its preferred over using index and name is unique here.

src/pages/calendar/components/Section.tsx
```diff
@@ -16,6 +18,33 @@ import { useDarkMode } from 'lib/hooks/useDarkMode';
...
+
+var names = {
+  'Reserved for BSENG students': 'SENG Only',
+  'Computer Science program.': 'CSC Only',
+  'BEng students': 'ENGR Only',
+  'Faculty of Engineering': 'ENGR/CSC Only',
+  'SCIENCE students': 'SCI only',
+  'BEng and BSEng students': 'BEng/BSeng Only',
+  'BME, BSEN, CENG, ELEC students': 'BME/SENG/CENG/ELEC Only',
+};
```
aomi, line 34: It's unclear from just this to understand what `names` is. If you can't express it in the data structure it might be good to rethink the data structure you're using or see if a comment is needed here.

aomi: Also, setting it as `var` has a lot of implications. First thing is that you're saying that its going to be modified but in the code there isn't any modifying so it can be a `const`. If you can make a stronger contract about what this value is used for and how you should take it.

In terms of `var` however it's a lot more complicated. You can read about it underneath but in general don't use it unless you are finding a specific need for it.
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/var
In the majority of cases a `let` or `const` will suffice.

src/pages/calendar/components/Section.tsx
```diff
@@ -59,9 +88,39 @@ export function SectionInfo({
...
+        name: 'Asynchronous',
+        color: 'cyan',
+      });
+    } else {
+      badges.push({
+        name: 'Synchronous',
+        color: 'cyan',
+      });
+    }
+  }
```
aomi, line 114: Nesting if statements is a bit of a "code-smell" (or _sus_ in modern terms perhaps).
I would either move this into a function or use some guard clauses to clean this up. 
From line 91 to 123 could be moved into a separate function it would help. This also makes testing a lot easier as it will separate your logic from the frontend code that displays the data.

src/pages/calendar/components/Section.tsx
aomi: In addition, favour code readability over small and perhaps hard to understand variable names too. In terms of performance it gets "transpiled" (which is the process of transforming code into another form of code) anyway so it doesn't really matter.

aomi (approved):

PR #274, opened 2021-08-22
feat: Booklist page

keithradford:
# Description

Adds the booklist feature for the Fall 2021 semester.  UI and code to be improved upon in the future, but for now we want this feature to be released to make sure users have access during the early weeks of the semester. Also so we can get amazon affiliate links.


src/pages/booklist/api/getCourseTextbooks.ts
```diff
@@ -0,0 +1,18 @@
...
+export type TextbookInfo = {
+  subject: string;
+  code: string;
+  term: string;
+  sections: Omit<CourseTextbooks, 'subject' | 'code'>[];
+};
+
+export const getCourseTextbooks = async (subject: string, code: string): Promise<TextbookInfo | null> => {
+  try {
+    const textbooks = await fetch(`/api/textbooks/202109/${subject}/${code}`);
```
aomi, line 12: We shouldn't hardcode this.

src/pages/booklist/hooks/useTextbooks.tsx
```diff
@@ -0,0 +1,41 @@
...
+import { useSavedCourses } from 'lib/hooks/useSavedCourses';
+
+import { getCourseTextbooks, TextbookInfo } from '../api/getCourseTextbooks';
+
+type TextbookResult = { status: 'loaded'; textbookInfo: TextbookInfo[] } | { status: 'loading' };
+
+export const useTextbooks = (): TextbookResult => {
+  const [result, setResult] = useState<TextbookResult>({ status: 'loading' });
+  const { courses } = useSavedCourses();
+  const { term } = useParams();
```
aomi, line 14: I wouldn't pass in term like this as we get a hard dependency on using this hook

aomi (approved):

PR #278, opened 2021-09-01
feat: API Rate Limiting

aomi:
# Description
This PR adds rate-limiting to the Cloud Function API endpoint. The rational for this endpoint is prevent abuse from upcoming endpoints that are publicly facing (anonymous auth required) and perform some sort of uncachable operation like a `POST` or `PUT` etc. 

The rate-limiting is accomplished through a middleware constructed using https://github.com/animir/node-rate-limiter-flexible, which is configured to use Redis. The Redis will be hosted on Redis Cloud in the same GCP region as the app.

For the time being, the rate-limiter isn't loaded in when developing locally so no Redis instance is needed to develop.

Rate-limiting is done on a per-IP basis. The default configuration is 10 requests a second. Once a user exceeds this, it will return a `429` and block them.

The rate-limiter can be adjusted using environment variables.
- `RATE_LIMIT_ENABLED`: if equal to `true`, will enable rate-limiter **if** the environment variable `REDIS_URI` is also defined. If it can't connect to Redis (the URI is needed) then it's not possible to start the rate-limiter.
- `RATE_LIMIT_POINTS`: Adjusts how many requests/second is permitted. Defaults to `10` unless this environment variable is passed in.

**Ideally**, we don't bloat this Cloud Function and _hopefully_ migrate this to the CDN/WAF level :cloud: :zap: 

## Checklist

- [ ] The code is documented.

## General Comments
This is the first usage of Redis in the app so it might not be as trivial to get things going. We should take it slow (rate-limit ourselves) when deploying and keep an eye on vitals.


functions/src/middlewares/rateLimiter.ts
```diff
@@ -0,0 +1,33 @@
...
+  duration: 1, // per 1 second by IP
+};
+
+const rateLimiter = new RateLimiterRedis(options);
+
+export const rateLimiterMiddleware = (
+  req: Request,
+  res: Response,
+  next: NextFunction
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any
```
keithradford, line 23: any way this can be typed?

keithradford (approved):
LGTM, 1 comment

functions/src/middlewares/rateLimiter.ts
aomi: a think a `void` might be fine looking back at this.

PR #280, opened 2021-09-09
feat: Return pre/co-requisite information in the backend

keithradford:
# Description

Adds the `preAndCoequisites` and `preOrCorequisites` fields to the `Course` model. This information is scraped and processed in course scraper, just needs to be piped through the backend so we can use it in the frontend.

## General Comments

This is branched off #252 so once that is merged this PR will be significantly smaller


JoshoTheMosho (approved):

aomi:
Going to merge this for the sake of seeing if its possible to hack in some pre-req info. Will leave on staging for now and do testing

PR #281, opened 2021-09-13
Bug Fix for issue #215 info button

### LINKED ISSUE #215: Info button doesn't work when searching course on timetable page
When searching for a course on [the timetable page](https://courseup.vikelabs.ca/scheduler/202109), clicking the "info" button (blue button with an `i`) does not work as expected.
It should take you to the course info page for that given course, but it currently does not.

Current info about the bug:

- The PID for the selected course is appended as a query param which is correct, however it is added to the URL with `scheduler` instead of `calendar`
- It appears to briefly redirect to the correct URL in the browser search bar but the page does not change

chendumpling:
# Description

Fixes a bug that causes a click event on the info button on timeline search results to have a conflict with another click event. The click event on the button redirects the page to the course information page as intended.

Closes #215

## Screenshots

This bug and fix only applies to the course cards that appear in the search results of the Timeline page.

Clicking the blue info icon as shown in this image...
image

### Before

...briefly shows the correct link address but is quickly replaced by an incorrect link address that also blocks the page from redirecting incorrectly.
image

### After

...now is no longer blocked and redirects to the correct link address.
image

### DIFF
--- a/src/common/sidebar/components/SearchResults.tsx
+++ b/src/common/sidebar/components/SearchResults.tsx
@@ -22,7 +22,7 @@ const SearchResults = ({ hits }: Props) => {
 
   const handleClick = (e: MouseEvent<HTMLDivElement>) => {
     const pid = e.currentTarget.getAttribute('data-pid');
-    if (pid && scheduleMatch === null) {
+    if (pid && !scheduleMatch) {
       setSearchParams({ pid });
     }
   };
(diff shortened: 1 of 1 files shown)


src/common/sidebar/components/SearchResults.tsx
```diff
@@ -22,7 +22,7 @@ const SearchResults = ({ hits }: Props) => {
 
   const handleClick = (e: MouseEvent<HTMLDivElement>) => {
     const pid = e.currentTarget.getAttribute('data-pid');
-    if (pid) {
+    if (pid && scheduleMatch === null) {
```
keithradford, line 25:
```suggestion
    if (pid && !scheduleMatch) {
```
Just a small simplification

keithradford (approved):
LGTM, thank you for doing this!

PR #283, opened 2021-09-15
fix: Prevent week view reset when course sections dropdown is triggered

### LINKED ISSUE #279: Bug: Course `More Information` button resets week view to current week
**Describe the Bug/Issue**
When a user clicks on the `More Information` button in the `Saved Courses` sidebar, the week viewed is reset to the current week.

**To Reproduce**
Steps to reproduce the bug:
1. Add a course to your CourseUp schedule
2. Change the week viewed
3. Click on the `More Information` button
4. Observe the week has been reset to the current week

**Expected Behaviour**
When expanding to gain more information on a course/sections, the currently viewed week should not be reset.

**Screenshots**
Added course and looking at a date in the future:
image
Clicked on the course `More Information` button and is back to present day:
image

**Desktop/Mobile**

 - OS: Windows 10
 - Browser: Firefox
 - Version: 91.0.2

logankuzyk:
Closes #279

Because showSections was a property of each course saved in LocalStorage, when the dropdown menu was opened, the stored course data would change, triggering a re-render of the whole calendar and resetting the week view.

I removed the showSections property of the saved courses, replacing it with a state variable in the CourseCard component. I also added a useEffect hook to make it so the sections are shown when the course is added and to collapse it when the course is removed from the timetable.

Because the showSections variable is no longer in the context of the SchedulerSidebar, I moved the SectionsCardContainer to be a child of CourseCard.

After I did this, I realized that in CourseCard, the checkbox's onChange event was getting fired when the sections menu was toggled, or when clicking anywhere on the card. I finally realized that the label prop of a parent Flex was the issue, so I removed it. I also removed the `cursor="pointer"` prop to show the users that clicking on the card itself does nothing.

## Checklist

- [ ] The code is documented.

### DIFF
--- a/src/lib/hooks/useSavedCourses.ts
+++ b/src/lib/hooks/useSavedCourses.ts
@@ -8,7 +8,7 @@ import { SECTION_TYPES, getFirstSectionType, hasSectionType } from 'lib/utils';
 
 import useLocalStorage from './storage/useLocalStorage';
 
-const OMITTED_FIELDS = ['lecture', 'lab', 'tutorial', 'color', 'sections', 'selected', 'showSections'];
+const OMITTED_FIELDS = ['lecture', 'lab', 'tutorial', 'color', 'sections', 'selected'];
 
 export const COLORS = [
   '#F56565', // red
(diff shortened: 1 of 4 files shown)


aomi:
@logankuzyk I'm noticing that some functionality is broken like switching sections.

PR #292, opened 2021-10-10
fix: update fetchers for timetable endpoint

keithradford:
forgot to run after making changes and merging


keithradford:
dont need to do this

PR #298, opened 2021-11-02
fix(docs): backend readme to be up to date

keithradford:
# Description

This README is a bit out of date, referencing NPM scripts we don't have etc. Just cleaning it up so new members don't get confused.

## General Comments
LMK if any descriptions are off or there's a step missing. I was having trouble reproducing on my machines 😬


aomi:
@VikeLabs/courseup-backend 
I'll defer the approval to the new people on the team. If it makes sense to the new members its likely :100: If it doesn't we need to fix our README.

kthlemaire:
Looks good to me! I am able to understand all of the instructions clearly. Thank you again Keith for updating it.

kthlemaire (approved):

PR #305, opened 2022-01-07
feat: allow loading in other terms for textbooks

aomi:
# Description
A simple PR to allow the term to be passed in as an argument rather than be hardcoded to a specific term. The GitHub action will now have a term input option.

The button itself will need to be enabled after we can verify the data is loaded in properly. This can be bypassed by just going directly to the page.

## Checklist

- [ ] The code is documented.


keithradford (approved):
LGTM

PR #313, opened 2022-01-30
feat: Refactor pages to use single Page component

### LINKED ISSUE #183: Bug: Search Function - Landing Page
**Describe the bug**

The search function does not work on the landing page.

**To Reproduce**

1. Go to https://courseup.vikelabs.ca
2. Click on search bar
3. Start searching for course
4. Notice no search being conducted
5. Switching the `courses` page results in search bar being reset

**Expected behavior**

Either the search bar is not present on the Landing Page (ideal situation + plus user get started prompt instead) or the search bar brings the user to the `courses` page and the search bar is not reset. 

**Screenshots**

Capture d’écran, le 2021-06-05 à 09 56 27

 - OS: macOS
 - Browser: Safari
 - Version: Latest

### LINKED ISSUE #312: Create Page component
We have a growing amount of different pages on CourseUp, and since we consistently have the header, feedback button, and sometimes sidebar it can be tedious to fill in the rest of the page with the actual content.

It would be beneficial if we had a `Page` component that the highest level of each page could be wrapped in so this formatting logic didn't have to be repeated, would minimize error and make refactors easy, as well as make new page creation easier in general.

For example [this chunk](https://github.com/VikeLabs/courseup/blob/main/src/pages/registration/containers/RegistrationContainer.tsx#L22-L28) of components on the registration page would be a single `Page` component.

- [ ] Find all instances of a page in our codebase
- [ ] Create consistent wrapper component (`Page` component) in our common folder
- [ ] Replace all pages with this component

keithradford:
# Description

Creates a general page component which can be used as a wrapper for any pages. Includes the header, optional sidebar, and title of the page. Makes it generally easier to create new pages and removes repeated logic.

Closes #312
closes #183 

## Checklist

- [ ] The code is documented.

## General Comments

todo: add tests


JoshoTheMosho (approved):
These changes all look good to me

src/common/page/Page.tsx
```diff
@@ -0,0 +1,65 @@
...
+
+  return (
+    <Flex h="100vh" direction="column" overflowX="hidden" overflowY="hidden">
+      <Helmet>
+        <title>{title}</title>
+      </Helmet>
+      <Header onSearchChange={handleSearchChange} />
+      {hasSidebar ? (
+        <Flex grow={1} overflow="hidden">
+          <ContentSidebar term={term as Term} searchQuery={query} />
```
aomi, line 53: I have a feeling that we shouldn't make the sidebar with a page by default a search one? This is the "SearchSidebar". It's not a generic sidebar container. The developer won't be able to configure this page component to use some other content within the sidebar area and they'll make a new component. 

Page > PageWithSidebar
PageWithSidebar > PageWithCoursesSidebar > Scheduler 
PageWithSidebar > TimetableSharing

src/common/page/Page.tsx
```diff
@@ -0,0 +1,65 @@
...
+    setQuery(q);
+  };
+
+  useEffect(() => {
+    if (term) {
+      setSavedTerm(term);
+    } else if (route) {
+      navigate(`/${route}/${savedTerm}`, { replace: true });
+    }
+  }, [navigate, route, savedTerm, setSavedTerm, term]);
```
aomi, line 43: If this logic is common, we could just make this itself a hook or something

src/common/page/Page.tsx
aomi: If we can make components like the SimpleGrid components from Chakra or the Modal components, it might help keep the code more composable and reusable.

aomi (approved):

PR #316, opened 2022-02-06
chore: add camelCase eslint rule as an error

aomi:
# Description
Given we are enforcing it via PRs but not actually with eslint, this adds camelCase as a rule which is considered an error if it fails.

https://eslint.org/docs/2.0.0/rules/camelcase

### DIFF
--- a/.eslintrc
+++ b/.eslintrc
@@ -47,6 +47,7 @@
         }
       }
     ],
+    "camelcase": [2, { "properties": "always" }],
     "import/no-unresolved": "off",
     "no-unused-vars": "off",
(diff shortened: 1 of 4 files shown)


keithradford:
Only related because of eslint, but is it possible to config so it fixes the import error on `npm run lint:fix` or if you have format on save enabled?

aomi:
@keithradford does it not? It _should_ afaik.

keithradford:
Maybe `npm run lint:fix` works but my format on save doesn't and it makes me sad :(

keithradford (approved):

PR #324, opened 2022-02-11
feat: Page component

keithradford:
See https://github.com/VikeLabs/courseup/pull/313 for full description


aomi (commented):
nit: perhaps we store page under `layouts` rather than `page/page`

keithradford:
+1, can store sidebar in there too

aomi (approved):

PR #334, opened 2022-02-16
feat: Next.js bootstrapping

aomi:
# Description
This PR bootstraps Next.js within the repository. It actually doesn't get used and all the commands that work for Next.js are documented in the README.

This PR also updates some core dependencies which have minor API changes of which have been properly addressed. The PR should not change the behaviour of the application and any such changes are errors/bugs introduced in this PR.

Subsequent work will start using Next.js and configure the deployment for it etc.

## Checklist

- [x] The code is documented.

### DIFF
--- a/src/pages/import/index.tsx
+++ b/src/pages/import/index.tsx
@@ -71,7 +71,7 @@ export function ImportTimetable(): JSX.Element {
   return (
     <Page title="View Timetable" leftSidebar={left} rightSidebar={right} mobileSupport>
       {isMobile ? (
-        <VStack pt={2} w="100%" h="100%" overflow="hidden" grow={1} px="3">
+        <VStack pt={2} w="100%" h="100%" overflow="hidden" flexGrow={1} px="3">
           <HStack justify="space-between" w="100%">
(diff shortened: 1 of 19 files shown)


aomi:
@szeckirjr added you explicitly as a reviewer as I had to update some props within the shared timetable code and want to check with you about it.

src/pages/import/index.tsx
```diff
@@ -87,7 +87,7 @@ export function ImportTimetable(): JSX.Element {
           {!loading && data ? calendarView ? calendarComponent : listView : <Spinner size="xl" />}
         </VStack>
       ) : (
-        <VStack pt={2} w="100%" height="100%" overflow="hidden" grow={1}>
+        <VStack pt={2} w="100%" height="100%" overflow="hidden" flexGrow={1}>
```
keithradford, line 90: is there a reason for prefixing these props with `flex`?

keithradford: i thought chakra added these  "sub-`flex`" props to flex components to simplify the prop names

aomi: Updating removed those props so I had to change them.

keithradford (approved):
I know we aren't using them yet but is it expected most the `next` npm commands fail?

PR #337, opened 2022-02-23
feat: Mobile calendar visual fixes

keithradford:
# Description

Fixes a visual bug where the times aren't visible on mobile calendars in light mode. Weirdest bug ever. Also made the toolbar responsive, so the calendar is usable in a read only mode. Enabled the calendar to be viewable from timetable sharing!

### Before

Screen Shot 2022-02-22 at 8 18 51 PM

### After

Screen Shot 2022-02-22 at 8 18 32 PM

## Checklist

- [ ] The code is documented.

### DIFF
--- a/src/pages/scheduler/components/SchedulerCalendar.tsx
+++ b/src/pages/scheduler/components/SchedulerCalendar.tsx
@@ -11,7 +11,7 @@ import { useDarkMode } from 'lib/hooks/useDarkMode';
 import { getCurrentTerm } from 'lib/utils/terms';
 
 import { CalendarEvent } from 'pages/scheduler/components/Event';
-import { CalendarToolBar } from 'pages/scheduler/components/Toolbar';
+import { CalendarToolBar, MobileToolBar } from 'pages/scheduler/components/Toolbar';
 import { useInitialDateTime } from 'pages/scheduler/hooks/useInitialDatetime';
 import { coursesToVCalendar } from 'pages/scheduler/shared/exporter';
 import { courseCalEventsToCustomEvents } from 'pages/scheduler/shared/transformers';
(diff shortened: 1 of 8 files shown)


src/pages/scheduler/components/SchedulerCalendar.tsx
```diff
@@ -83,7 +83,9 @@ export const SchedulerCalendar = ({ term, courseCalendarEvents = [] }: Scheduler
       eventPropGetter={eventPropGetter}
       slotPropGetter={slotPropGetter(mode)}
       components={{
-        toolbar: CalendarToolBar(setSelectedDate, term || getCurrentTerm(), vCalendar),
+        toolbar: isMobile
+          ? MobileToolBar(setSelectedDate, term || getCurrentTerm(), vCalendar)
+          : CalendarToolBar(setSelectedDate, term || getCurrentTerm(), vCalendar),
```
aomi, line 88: Could the `term` value be defined above to be defined. I think it'll require some object destruct renaming but it would clean this part up a lot. If we always default back to `getCurrentTerm` you could set the default value of that prop with it. 

Not an issue introduced from this PR but it was emphasized so just pointing it out.

src/pages/scheduler/components/Toolbar.tsx
```diff
@@ -7,6 +7,69 @@ import { logEvent } from 'lib/utils/logEvent';
...
+            aria-label="Next day"
+            colorScheme="gray"
+            icon={<ChevronRightIcon />}
+            size="sm"
+            onClick={handleClick(1)}
+          />
+        </HStack>
+      </Flex>
+    );
+  };
```
aomi, line 71: Do they not share enough code to make it one component? I feel this adds a lot of duplicated code.

keithradford: yaaaa I got lazy

aomi (approved):

PR #340, opened 2022-02-28
feat: Events endpoint

aomi:
Fixes tsoa by hoisting the types from the library. Adds a `events` collection and corresponding endpoint.


functions/src/events/Event.controller.ts
```diff
@@ -0,0 +1,27 @@
...
+import { add } from 'typesaurus';
+import { EventsCollection } from '../db/collections';
+
+@Route('events')
+export class EventsController extends Controller {
+  /**
+   * Retrieves all the courses available. If query params are passed in, they will be used to filter results.
+   * @param subject
+   * @param code
+   */
```
keithradford, line 11: 👀

functions/src/events/Event.controller.ts
```diff
@@ -0,0 +1,27 @@
...
+  @Post()
+  public async postEvent(@Query() id: string): Promise<{ success: boolean }> {
```
keithradford, line 13: is `id` the best name? Not sure what you had in mind for intended use but if we're passing something like `Contest page visit` or something perhaps just `event` is better

keithradford (dismissed):

keithradford:
oops didn't mean to approve

functions/src/events/Event.controller.ts
aomi: `id` just maps to what the URLs query params are where `id` is actually the location. I'll change it .

functions/src/events/Event.model.ts
```diff
@@ -0,0 +1,4 @@
+export type EventRequest = {
+  name: string;
+  [key: string]: any;
```
keithradford, line 3: I guess `any` is to keep it flexible with what we pass in? Should prob just add an `eslint-ignore` even though it's just a warning

keithradford (approved):

PR #356, opened 2022-03-16
filter dropdown is now open permantently;

### LINKED ISSUE #350: Make filter dropdown permanently open
I think as long as we have a minimal amount of filtering options (which we do) it's better visibility to leave it open permanently. Also, should we maybe store the value in session storage?

fluffyponyy:
# Description

The filters dropdown is now permanently open, displaying the one filter currently in place. The 'Filters' button which opened the dropdown is now completely deleted, since it is unnecessary. 

Closes #350 

### Before
image

### After
image

## Checklist

- [X] The code follows all style guidelines.
- [ ] The code is documented.
- [X] I have self-reviewed my changes and have done QA.

## General Comments

I am not sure how to add the state to session storage, and I am also unsure of the style guidelines or how code is tested+documented with CourseUp.

### DIFF
--- a/src/common/layouts/sidebar/variants/Courses.tsx
+++ b/src/common/layouts/sidebar/variants/Courses.tsx
@@ -90,7 +90,11 @@ export function CoursesTopBar({ onFilter }: TopBarProps): JSX.Element {
             <FormLabel htmlFor="email-alerts" mb="0" fontSize="sm">
               Only Show Courses in Session
             </FormLabel>
-            <Switch id="email-alerts" onChange={(e) => onFilter && onFilter(e.currentTarget.checked)} isChecked={filter} />
+            <Switch
+              id="email-alerts"
+              onChange={(e) => onFilter && onFilter(e.currentTarget.checked)}
+              isChecked={filter}
+            />
           </Flex>
         </FormControl>
       </Box>
@@ -107,7 +111,6 @@ type Props = {
 };
 
 export function Courses({ term }: Props): JSX.Element | null {
-  //const [filter, setFilter] = useState(false);
   const [filter, setFilter] = useSessionStorage<boolean>('filter_courses', true);
   const { data: subjects, loading: subjectsLoading } = useSubjects({ term: term as Term });
   const { data: courses, loading: coursesLoading } = useGetCourses({
--- a/src/common/layouts/sidebar/variants/Courses.tsx
+++ b/src/common/layouts/sidebar/variants/Courses.tsx
@@ -1,4 +1,4 @@
-import { useMemo, useState } from 'react';
+import { useMemo } from 'react';
 
 import { ChevronRightIcon } from '@chakra-ui/icons';
 import {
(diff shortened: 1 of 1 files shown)


aomi:
@fluffyponyy thanks for the PR! Addressing some of your general comments here.
There's some basics covered in https://github.com/VikeLabs/courseup/wiki/Theme-Guide about "style-guide" and what not. Anything not in there we'll chime in or answer in the Discord as needed.

To use session storage, I would suggest importing https://github.com/VikeLabs/courseup/blob/main/src/lib/hooks/storage/useSessionStorage.ts

You would then use it something like
```tsx
function MyComponent() {
  const [filterValue, setFilterValue] = useSessionStorage<boolean>('filter_courses', true)
}
```

src/common/layouts/sidebar/variants/Courses.tsx
```diff
@@ -116,7 +107,8 @@ type Props = {
 };
 
 export function Courses({ term }: Props): JSX.Element | null {
-  const [filter, setFilter] = useState(false);
+  //const [filter, setFilter] = useState(false);
```
keithradford, line 110:
```suggestion
```

keithradford (commented):
Thank you for the contribution!

src/common/layouts/sidebar/variants/Courses.tsx
```diff
@@ -51,11 +49,11 @@ export interface TopBarProps {
 }
 
 export function CoursesTopBar({ onFilter }: TopBarProps): JSX.Element {
-  const { isOpen, onToggle } = useDisclosure();
   const { term } = useParams();
   const location = useLocation();
   const [searchParams] = useSearchParams();
   const mode = useDarkMode();
+  const [filter] = useSessionStorage<boolean>('filter_courses', true);
```
aomi, line 56: I misguided you :bow: 
You'll want to remove this and add a property called `filter` on `TopBarProps` here instead.
Then you'll want to pass the `filter` value into `<CourseTopBar/>` on line 134. 

That way we avoid having multiple uses of hooks and the switch button remains in sync with the session storage value :)

aomi: While this would have worked the same, this is a bit of a cleaner way.

keithradford:
Hey @fluffyponyy any update on this PR? No rush, just wanted to check in

src/common/layouts/sidebar/variants/Courses.tsx
```diff
@@ -136,7 +131,7 @@ export function Courses({ term }: Props): JSX.Element | null {
 
   return (
     <>
-      <CoursesTopBar onFilter={handleFilter} />
+      <CoursesTopBar onFilter={handleFilter} filter/>
```
keithradford, line 134: I'm curious, does this work as expected? Since `filter` is a boolean in the props type, my understanding is naming the prop as you've done here will mean it's always true. Doing `filter={filter}` would correctly pass the variable that can be true or false.
I'm not 100% sure I'm right though so if this works just let me know

keithradford (commented):
You can run `npm run lint:fix` to automatically fix the linting errors you have 👍

PR #366, opened 2022-04-27
feat: Not Found Component

### LINKED ISSUE #310: Improve "Unable to find saved courses" component
If users do not have any courses saved when attempting to register or look at their booklist, the message: "Unable to find saved courses for <Term>" appears. This can be somewhat vague, given users might not know "saved courses" are what appear on their timetables. Furthermore, we are just leaving users on their own to go add these.

The message should be re-worded to be more clear and provide a link/button to the timetable view for users to click and add courses.

TODO:
- [ ] Increase clarity of sentence
- [ ] Create link/button to timetable view
- [ ] make this it's own component since it is reused in different places
- [ ] write unit tests

Current view:

liamshatzel:
# Description

Replaces hardcoded text/button/term with a single component. Specifically on explore courses page, registration and booklist (booklist has two cases: no courses in the timetable, no textbooks found for the courses in the timetable).

Closes #310 

### Before

Booklist: 
Screen Shot 2022-04-26 at 7 47 32 PM

Registration:
Screen Shot 2022-04-26 at 7 55 37 PM

### After

No saved courses:
Screen Shot 2022-04-26 at 8 00 40 PM

No textbooks found:
Screen Shot 2022-04-26 at 7 48 51 PM

Registration After:
Screen Shot 2022-04-26 at 7 56 06 PM

## Checklist

- [ ] The code is documented.


src/pages/booklist/containers/BooklistContainer.tsx
```diff
@@ -51,18 +48,16 @@ export function BooklistContainer(): JSX.Element | null {
               .map(({ sections, subject, code }) => {
                 return <TextbookCard subject={subject} code={code} sections={sections} />;
               })
+          ) : textbooks.textbookInfo.filter((textbook) => textbook && textbook.term === term).length <= 0 &&
+            textbooks.textbookInfo.length <= 0 ? (
+            <NotFound item="Unable to find saved courses from your" term={term} timetableButton timetable />
```
keithradford, line 53: Is there a case where `timetable != timetableButton`? I think these can probably be merged to one prop `timetable`

liamshatzel: Good point, just made the change 👍. I don't think there will be a case in the future when they don't go together either.

src/common/notfound/NotFound.tsx
```diff
@@ -0,0 +1,49 @@
...
+import { getReadableTerm } from 'lib/utils/terms';
+
+type Props = {
+  item: string;
+  term: string;
+  timetableButton?: boolean;
+  timetable?: boolean;
+};
+
+export function NotFound({ item, term, timetable }: Props) {
```
keithradford, line 16:
```suggestion
export function NotFound({ children, term, timetable }: PropsWithChildren<Props>) {
```
Let's replace `item` with `children` using `PropsWithChildren`. There are a few examples in the codebase that use this, but essentially this will allow us to use the component like:
```jsx
<NotFound term="202109">
   Unable to find saved courses from your
</NotFound>
```

src/common/notfound/NotFound.tsx
```diff
@@ -0,0 +1,49 @@
...
+  const mode = useDarkMode();
+  var display: string = '';
+  if (timetable) display = ' timetable';
```
keithradford, line 19: Not sure I'm a fan of this. I think we should just have the full string passed in in `children` (what is currently `item`) and only conditionally render the button based on the prop

src/common/notfound/NotFound.tsx
```diff
@@ -0,0 +1,49 @@
...
+  return (
+    <Container maxW="container.xl" centerContent data-testid="courses-not-found">
+      <Divider my="4" />
+      <Box padding="10">
+        <Flex direction={{ md: 'row', base: 'column' }} justifyContent="space-between">
+          <Heading size="md" color={mode('gray', 'dark.header')}>
+            {`${item} `}
```
keithradford, line 26:
```suggestion
            {item}
```
since it is defined as a string this should work

src/pages/booklist/containers/BooklistContainer.tsx
```diff
@@ -1,26 +1,23 @@
...
 import { Center, Spinner, Text, VStack } from '@chakra-ui/react';
 import { useParams } from 'react-router';
 
 import { Term } from 'lib/fetchers';
-import { useDarkMode } from 'lib/hooks/useDarkMode';
 import { logEvent } from 'lib/utils/logEvent';
 import { getReadableTerm } from 'lib/utils/terms';
 
 import { Page } from 'common/layouts/Page';
+import { NotFound } from 'common/notfound/NotFound';
```
keithradford, line 12:
```suggestion
import { NotFound } from 'common/notFound/NotFound';
```
what I mean by this is renaming the folder, not just the change in the code

src/pages/booklist/containers/BooklistContainer.tsx
```diff
@@ -51,18 +48,11 @@ export function BooklistContainer(): JSX.Element | null {
...
-              <Container alignItems="center" maxW="container.xl">
-                <Heading size="md" color={mode('gray', 'dark.header')}>
-                  Unable to find saved courses or textbooks for{' '}
-                  <Text as="span" color={mode('black', 'white')}>
-                    {getReadableTerm(term)}
-                  </Text>
-                </Heading>
-              </Container>
-            </>
+            <NotFound item="No textbooks found for your saved courses in" term={term} timetable={false} />
```
keithradford, line 55:
```suggestion
            <NotFound item="No textbooks found for your saved courses in" term={term} />
```
since `timetable` is optional (uses the `?` in the type definition) not passing it will count it as falsey, which is generally a bit more clean

src/pages/scheduler/components/share/SelectedCoursesCardList.tsx
```diff
@@ -43,9 +43,7 @@ export function SelectedCoursesCardList({ courses, term }: SelectedCoursesTableP
           return null;
         })
       ) : (
-        <Text>
-          Unable to find saved courses for <Text as="strong"> {getReadableTerm(term || '')} </Text>
-        </Text>
+        <NotFound item="Unable to find saved courses for" term={term} timetable={false} />
```
keithradford, line 46:
```suggestion
        <NotFound item="Unable to find saved courses for" term={term} />
```

src/common/notfound/NotFound.tsx
liamshatzel: Would it be best to change the wording so that the term comes at the end of the string? Its just a bit more wordy
Screen Shot 2022-05-12 at 9 15 38 AM
.

PR #373, opened 2022-05-14
feat: Allow using Banner v9 directly for existing backend solution

aomi:
# Description
This is a temporary solution to allow the existing backend to return data accessing Banner v9 instead of Banner v8 (through the uvic-course-scraper). Once the new backend is implemented this will be removed and deprecated ASAP. 

The rational for this implementation approach is that this involves modifying the minimal amount of modules and could be implemented in time before Banner v8 is shut down.

**Note**: Frontend React code does not call the v9 version yet. From the `useSections` hook, a query param of `v9` set to `true` will make the request process through Banner v9.
## Checklist

- [ ] The code is documented.


functions/src/banner/banner.ts
```diff
@@ -0,0 +1,87 @@
+export interface Section {
+    id: number;
+    term: string;
```
keithradford, line 3: linting

functions/src/banner/bannerClient.ts
```diff
@@ -0,0 +1,234 @@
...
+  saveCookie(response: Response<string>, term: string): void {
+    // get the cookies
+    const cookies = response.headers['set-cookie'];
+    // if there's no cookies, return
+    if (cookies) {
+      this.cookies = { ...this.cookies, [term]: cookies };
+    }
+  }
+
+  async init() {
```
keithradford, line 79:
```suggestion
  async init(): void {
```
Prob do this for the rest of the methods too

functions/src/banner/bannerClient.ts
```diff
@@ -0,0 +1,234 @@
...
+    const urlParams = new URLSearchParams({
+      searchTerms: '',
+      offset: offset.toString(),
+      max: max.toString(),
+    });
+
+    const url = `${BANNER_SSB_GET_TERMS}?${urlParams}`;
+
+    const response = await got(url).json<TermsResponse[]>();
+    // console.log(response)
```
keithradford, line 110:
```suggestion
```

keithradford (approved):
Looks good for now once linting fixed

PR #374, opened 2022-05-16
Update terms to support Winter 2022-23

keithradford:
# Description

Changes the frontend to support the new calendar for winter 22/23. Changed the banner to advertise that we support it cause why not.

## Screenshots

Screen Shot 2022-05-16 at 10 54 46 AM


aomi (approved):
Can't merge till all the other PRs are merged + the backend population scripts

PR #384, opened 2022-05-20
fix: Change buildingAccronym to buildingAbbreviation

keithradford:
# Description

Changes the sections endpoint to return `buildingAbbreviation` instead of `buildingAccronym` as abbreviation is a more accurate representation of the data and this matches what we have in the frontend.

## General Comments
I think `routes.ts` changes are all style based 🤷


aomi (approved):
LGTM

PR #386, opened 2022-05-22
fix: `meetingTime.where` to be `TBA`

keithradford:
# Description

Fixes the V9 endpoint to return `TBA` for `meetingTime.where` instead of `null null` to match the V8 data. This happens for online courses typically.

## Checklist

- [ ] The code is documented.

### DIFF
--- a/functions/src/sections/Section.controller.ts
+++ b/functions/src/sections/Section.controller.ts
@@ -49,14 +49,16 @@ export class SectionsController extends Controller {
                 formatTime(m.meetingTime.endTime)
               : 'TBA',
           days: days(m.meetingTime),
-          where: `${m.meetingTime.buildingDescription} ${m.meetingTime.room}`,
+          where: m.meetingTime.buildingDescription
+            ? `${m.meetingTime.buildingDescription} ${m.meetingTime.room}`
+            : 'TBA',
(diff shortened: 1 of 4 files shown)


functions/src/sections/Section.controller.ts
```diff
@@ -49,14 +49,16 @@ export class SectionsController extends Controller {
...
+            ? `${m.meetingTime.buildingDescription} ${m.meetingTime.room}`
+            : 'TBA',
           dateRange: `${m.meetingTime.startDate} - ${m.meetingTime.endDate}`,
           scheduleType: s.scheduleTypeDescription,
           instructors: s.faculty.map(
             (f) => f.displayName + `${f.primaryIndicator ? ' (P)' : ''}`
           ),
           building: m.meetingTime.buildingDescription ?? undefined,
-          buildingAccronym: Buildings.get(
+          buildingAbbreviation: Buildings.get(
```
szeckirjr, line 61: are we changing back to `buildingAbbreviation` instead of acronym?

keithradford: It's supposed to be abbreviation (this PR has the same change but in the v8 context https://github.com/VikeLabs/courseup/pull/384/files)

szeckirjr (approved):
with the little I know from how this works, LGTM 🎉

PR #388, opened 2022-05-26
Using building abbreviations 

### LINKED ISSUE #158: Compact location names on the timetable calendar and saved courses sidebar.
Currently the calendar event component and saved courses display the full length location of the meeting time as shown below:
 

The location names on these views should be compacted into their respective short forms like so:
- Elliott Building 162 to ELL 162
- Engineering Lab Wing B215 to ELW B215

The implementation would likely involve a lookup table which contains the full length building/location name as the key to the shorthand version. For example:
```js
// lookup table
const locations = {
'Engineering Lab Wing': 'ELW' 
};

// using lookup table
const shorthand = locations['Engineering Lab Wing']
```

If a list of location / buildings cannot be found online, ping @aomi @keithradford for guidance / assistance

szeckirjr:
# Description

This PR adds building abbreviations to CourseUp.
In the timetable sidebar and course page, full name is shown on larger screens and shorter name is shown in smaller screens.
In the timetable calendar, the shorter name is default.
All short names are Tooltips that when hovered show the full building name.

Closes #158 

# Screenshots

## Before
image
image
image

### Always short
image

### Short on small screens
image
image

### Short names are hoverable
image

## Checklist

- [ ] The code is documented.


src/pages/calendar/components/Schedule.tsx
```diff
@@ -34,7 +36,17 @@ export function Schedule({ meetingTimes }: ScheduleProps): JSX.Element {
             <Td>{m.time}</Td>
             {/* TODO: verify if we can safely exclude this for most cases */}
             {/* <Td>{m.scheduleType}</Td> */}
-            <Td>{m.where}</Td>
+            <Td>
+              {isMobile ? (
```
keithradford, line 40: Since this logic is repeated a bit, might be good to isolate it in it's own component.

Something like
```jsx
<Location short={`${m.buildingAbbreviation} ${m.roomNumber}`} long={m.where} />
```
For example. Could even have a prop like `alwaysShort` or something for the Event so it never uses the long form.

szeckirjr: that's a good point!! I just pushed some commits that should make that change

src/common/location/__tests__/Location.test.tsx
```diff
@@ -0,0 +1,56 @@
...
+import Location from '../Location';
+
+jest.mock('@chakra-ui/react', () => ({
+  ...jest.requireActual('@chakra-ui/react'),
+  useMediaQuery: jest.fn().mockReturnValue(false),
+}));
+
+describe('Location Component', () => {
+  it('should render the short form on mobile', () => {
+    (useMediaQuery as unknown as jest.Mock).mockReturnValue([true]);
```
keithradford, line 14: Just wondering why this type casting is required? Not dogging it cause I know testing can be a pain sometimes just curious.

src/common/location/Location.tsx
```diff
@@ -0,0 +1,20 @@
+import { Tooltip, useMediaQuery, Text } from '@chakra-ui/react';
+
+export default function Location({
+  short,
+  long,
+  alwaysShort,
+}: {
+  short?: String;
+  long: String;
+  alwaysShort: boolean;
```
keithradford, line 10:
```suggestion
  alwaysShort?: boolean;
```
Then you can not pass the prop when you want it to be false like:
```jsx
<Location short={...} long={...} />
```
and then something like this when you want it to be true:
```jsx
<Location short={...} long={...} alwaysShort />
```

keithradford (approved):
great tests 🙌 just a small style nit

src/common/location/__tests__/Location.test.tsx
szeckirjr: yeah I was having trouble mocking it so tried a bunch of different things but just tried again and the `as unknown` is not necessary, just casting it as jest Mock works as I need in that type to be able to mockReturnValue! I've simplified the type casting on it

keithradford:
feel free to merge whenever!

PR #391, opened 2022-06-05
Nishchint/custom color for courses

### LINKED ISSUE #361: Allow users to set custom colours for courses
We can still use our default colours, but I thought it might be nice to allow it to be customized

Screen Shot 2022-03-30 at 8 04 15 PM
I imagine it working like someone clicks the coloured portion of the card and a colour picker comes up

very low priority, might not be doable but might as well make an issue for it

NishchintDhawan:
# Description

This PR provides the user to select a custom color for the courses currently selected. The current implementation provides a way to select a color from the provided list of colors which comes up when the user clicks on the edit icon. 

Closes #361 

### After

When the user has clicked on the edit icon:
image

Dark mode:
image

## Checklist

- [ ] The code is documented.

## General Comments

The  user interaction is not finalised but just one of the ways to solve the issue. Another option could be to provide custom colors. The issue with that is if the user chooses white or black as colors, the edit button might not be visible along with the room number. Open for discussion !

### DIFF
--- a/src/lib/hooks/useSavedCourses.ts
+++ b/src/lib/hooks/useSavedCourses.ts
@@ -66,6 +66,7 @@ type SavedCourses = {
   clearCourses: (term: string) => void;
   setSection: (type: string, newSection: SavedSection, existingCourse: SavedCourse) => void;
   contains: (pid: string, term: string) => boolean;
+  updateCourseColor: (pid: string, term: string, color: string) => void;
   sectionIsSaved: (pid: string, term: string, sectionCode: string) => boolean;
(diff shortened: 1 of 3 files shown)


FlyteWizard:
> Another option could be to provide custom colours. The issue with that is if the user chooses white or black as colours, the edit button might not be visible along with the room number.

A possible solution for the future is `color-contrast()` but this is still a Safari experimental feature.

[https://developer.mozilla.org/en-US/docs/Web/CSS/color_value/color-contrast](https://developer.mozilla.org/en-US/docs/Web/CSS/color_value/color-contrast)

PR #401, opened 2022-06-24
Feat: Responsive timetable

### LINKED ISSUE #318: Make timetable page responsive

szeckirjr:
# Description

This PR changes the Page component to be responsive on mobile using Swiper. A menu bar at the bottom shows your current page and you can swipe or click to switch between them.

Closes #318 

### Before (no way of switching away from the scheduler screen)
image

### After
image

Here's a little GIF of it in action: https://imgur.com/a/wj0E4x6

## Checklist

- [X] The code passes all required tests.
- [ ] The code is documented.

## General comments

When a user starts typing something in the menubar, it slides them back to the left most slide (left sidebar, as it includes search results).
This is done using a bit of a workaround using `useState`. The best would be to use the `useSwiper` hook they provide, but that needs to be in a _direct_ child component of `Swiper`, which is not possible for our use case since the menu bar is the one that triggers the slide, which won't be inside Swiper...

### DIFF
--- a/jest.config.js
+++ b/jest.config.js
@@ -23,6 +23,7 @@ const config = {
   modulePaths: ['<rootDir>/src/'],
   moduleNameMapper: {
     '^.+\\.(css|scss)$': '<rootDir>/src/lib/utils/jest/__mocks__/styleMock.js',
+    '^swiper.*$': '<rootDir>/src/lib/utils/jest/__mocks__/moduleMock.js',
   },
 };
 
--- a/src/pages/scheduler/containers/SchedulerContainer.tsx
+++ b/src/pages/scheduler/containers/SchedulerContainer.tsx
@@ -4,6 +4,7 @@ import { Box, Flex } from '@chakra-ui/react';
 import { useParams } from 'react-router';
 
 import { useSavedCourses } from 'lib/hooks/useSavedCourses';
+import { useSmallScreen } from 'lib/hooks/useSmallScreen';
 import { getCurrentTerm } from 'lib/utils/terms';
 
 import { SchedulerCalendar } from '../components/SchedulerCalendar';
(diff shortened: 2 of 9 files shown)


jest.config.js
```diff
@@ -23,6 +23,7 @@ const config = {
   modulePaths: ['<rootDir>/src/'],
   moduleNameMapper: {
     '^.+\\.(css|scss)$': '<rootDir>/src/lib/utils/jest/__mocks__/styleMock.js',
+    '^swiper.*$': '<rootDir>/src/lib/utils/jest/__mocks__/moduleMock.js',
```
keithradford, line 26: Did you end up needing this?

src/pages/scheduler/containers/SchedulerContainer.tsx
```diff
@@ -22,10 +23,11 @@ export function SchedulerContainer(): JSX.Element {
     () => denormalizeCourseEvents(coursesResult.status === 'loaded' ? coursesResult.data : []),
     [coursesResult]
   );
+  const smallScreen = useSmallScreen();
 
   return (
     <Flex flexGrow={1} height="100%" overflow="hidden" direction={'column'}>
-      <Box w="100%" height="100%" px="3" py="2">
+      <Box w="100%" height="100%" px={smallScreen ? '0' : '3'} py={smallScreen ? '1' : '2'}>
```
keithradford, line 30: Can we maybe try px=1, pb=0, and py stays at 2 for small screens? I think it's a bit tight with no padding on the sides but just a lil would be good, and then it would be nice if it went all the way to the bottom.

keithradford (approved):
Would love to try this out on staging so I can play around with actually adding courses!

jest.config.js
szeckirjr: yep, tests fail if swiper module is not mocked :/

keithradford: no worries just wanted to make sure

PR #405, opened 2022-07-09
Changing style of import page

szeckirjr:
# Description

Proposal for fixing visual layout of import page on mobile to make use of sidebars and remove redundant buttons.
Removing right sidebar, as it simply showed list of courses, which is accessible on main page. Also added timetable tags to the left sidebar along with the actions.

### Before
image
image
image

### After
image
image
image

## Checklist

- [ ] The code is documented.

## General Comments

Would it be useful to add timetable cards underneath for "Your Timetable Courses" to show the user what they have saved currently?

### DIFF
--- a/src/pages/import/index.tsx
+++ b/src/pages/import/index.tsx
@@ -68,7 +68,7 @@ export function ImportTimetable(): JSX.Element {
   return (
     <Page title="View Timetable" leftSidebar={left} rightSidebar={right} mobileSupport>
       {smallScreen ? (
-        <VStack pt={2} w="100%" h="100%" overflow="hidden" flexGrow={1} px="3">
+        <VStack pt={2} w="100%" h="100%" overflow="hidden" flexGrow={1} px={1}>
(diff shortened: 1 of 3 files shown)


src/pages/import/index.tsx
```diff
@@ -14,44 +14,51 @@ import { Sidebar } from 'common/layouts/sidebar/containers/Sidebar';
...
 import { TimetableCourseCard } from './components/TimetableCourseCard';
+import { TimetableCourseTags } from './components/TimetableCourseTags';
 
 export function ImportTimetable(): JSX.Element {
   const { slug } = useParams();
   const smallScreen = useSmallScreen();
 
   const { loading, data } = useGetTimetable({ slug: slug });
+  const [courses, setCourses] = useState<TimetableCourse[]>([]);
+  const [term, setTerm] = useState<Term>('202205');
```
keithradford, line 25: `getCurrentTerm()` probably better here instead of hard coding

PR #416, opened 2022-10-08
Upsert scripts

szeckirjr:
# Description

Closes #XX

## Checklist

- [ ] The code is documented.


scripts/upsert/sections.ts
```diff
@@ -0,0 +1,53 @@
...
+
+if (process.env[2] && process.env[3]) {
+  const registrationDate = new Date(process.argv[2]);
+  const dropDate = new Date(process.argv[3]);
+
+  const terms = new Term().sessionTerms();
+
+  terms.forEach(async (term) => {
+    await upsertSectionsScript(term.toString(), registrationDate, dropDate);
+  });
```
aomi, line 52: I got called out for this in while developing our SENG 499 backend but this isn't what you want to do.
Change this into a normal loop. This will cause issues if it throws.

scripts/upsert/sections.ts
```diff
@@ -0,0 +1,53 @@
...
+    const { data } = await getSearchResults(fc, { term, max: init.pageMaxSize, offset });
+    data.concat(data);
+    sectionsCount += data.length;
+  }
+
+  await createTask(`upsertSections-${term}`, upsertSections(fc, term, data), {});
+};
+
+if (process.env[2] && process.env[3]) {
+  const registrationDate = new Date(process.argv[2]);
```
aomi, line 45: We probably want some default values?

scripts/upsert/sections.ts
```diff
@@ -0,0 +1,53 @@
...
+import { upsertSections } from '../../lib/sections';
+import { findLatestTask, createTask } from '../../lib/task';
+import { Fetch } from '../../lib/banner/fetch';
+import makeFetchCookie from 'fetch-cookie';
+import { getSearchResults } from '../../lib/banner';
+import { range } from '../../lib/fn';
+import fetch from 'node-fetch';
+
+export const upsertSectionsScript = async (term: string, registrationDay: Date, dropDate: Date) => {
+  const today = new Date();
```
aomi, line 12: Passing in `today` is an object would probably help avoid mocking timers?

scripts/upsert/sections.ts
```diff
@@ -0,0 +1,53 @@
+import { Term } from '../../lib/term';
+import { differenceInMinutes } from 'date-fns';
+import { upsertSections } from '../../lib/sections';
+import { findLatestTask, createTask } from '../../lib/task';
+import { Fetch } from '../../lib/banner/fetch';
+import makeFetchCookie from 'fetch-cookie';
+import { getSearchResults } from '../../lib/banner';
+import { range } from '../../lib/fn';
+import fetch from 'node-fetch';
```
aomi, line 9: Fetch is actually globally patched in for scripts so it makes Next.js runtime

scripts/upsert/courses.ts
```diff
@@ -0,0 +1,36 @@
...
+import { differenceInDays, differenceInMinutes } from 'date-fns';
+import { upsertCourses } from '../../lib/courses';
+import { findLatestTask, createTask } from '../../lib/task';
+
+export const upsertCoursesScript = async (term: string, registrationDay: Date, dropDate: Date) => {
+  const today = new Date();
+
+  const daysUntilRegistration = differenceInDays(registrationDay, today);
+
+  const lastUpdated = await findLatestTask(`upsertCourses-${term}`);
```
aomi, line 11: I almost feel making a `term` column here might be good for the tasks table so? What do you think?

scripts/upsert/courses.ts
```diff
@@ -0,0 +1,36 @@
...
+
+  const minutesSinceLastUpdate = differenceInMinutes(today, lastUpdated?.startedAt ?? -1);
+
+  // If over a week away from registration day, run task less frequently
+  if (daysUntilRegistration > 7) {
+    // If before drop date and courses have been updated within the last hour, return
+    if (today < dropDate && minutesSinceLastUpdate <= 60) return;
+    // If after drop date and courses have been updated within the last 24 hours, return
+    if (today > dropDate && minutesSinceLastUpdate <= 1440) return;
+  }
```
aomi, line 21: Is another case not needed to stop updating after the drop date?

scripts/upsert/courses.ts
```diff
@@ -0,0 +1,36 @@
...
+
+  // Else, run the task
+  await createTask(`upsertCourses-${term}`, upsertCourses(term), {});
+};
+
+if (process.env[2] && process.env[3]) {
+  const registrationDate = new Date(process.argv[2]);
+  const dropDate = new Date(process.argv[3]);
+
+  const terms = new Term().sessionTerms();
```
aomi, line 31: This will run into an issue where for summer session where it will need to become summer session to run at all. Same issue with adding winter session data during summer. It needs to look ahead I think.

scripts/upsert/sections.ts
szeckirjr: Turning it into a for loop and removing the forEach? I should still keep the await though right?

scripts/upsert/sections.ts
szeckirjr: Yeah I was thinking that and I had some at first, but then changed it so that the scripts will _only_ run if the arguments are provided because I wasn't sure which dates to pick (I assume most current term?). Let me know what you think and I can add defaults

scripts/upsert/courses.ts
szeckirjr: Hmm that might be useful... if tasks are mostly related to term it'd be helpful to specify that. But is there a downside for using the string params instead?

scripts/upsert/sections.ts
szeckirjr: image
I get this error if I remove that :/

scripts/upsert/courses.ts
szeckirjr: Sure - I was just under the impressions we always want updated data, just less frequent! What's a good business logic for this case then?
Within a week of registration -> every 30min
Before drop date -> every hour
After drop date -> stop?

PR #417, opened 2022-11-06
Fix seats not loading cause of problems with /seats endpoint

szeckirjr:
# Description

500 error response from /seats endpoint is causing none of the information for a course to show.

image

The seating data is currently coming through v9 object, so I just replaced the seats with a modified version of the data to match the expected type, and it seems to be working on my end.

I deleted the /seats config from swagger.json and regenerated `fetchers.tsx` with the `fetcher:generate` command.

## After
(No more /seats request)
image

## Checklist

- [X] The code follows all style guidelines.
- [X] The code passes all required tests.
- [ ] The code is documented.
- [X] I have self-reviewed my changes and have done QA.

### DIFF
--- a/functions/build/swagger.json
+++ b/functions/build/swagger.json
@@ -1,1333 +1,1189 @@
 {
-	"components": {
-		"examples": {},
-		"headers": {},
-		"parameters": {},
-		"requestBodies": {},
-		"responses": {},
-		"schemas": {
-			"Course": {
-				"properties": {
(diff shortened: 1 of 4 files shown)


aomi:
If you're going to remove the `/seats` endpoint its best to remove it from the `functions/src/sections/*.controller.ts` files. 
These controller files defined the spec file you've modified so next time the spec is generated your changes will be lost.

PR #421, opened 2022-11-12
Outage & Bug Warning

szeckirjr:
# Description

Adds some extra warnings to the CourseUp banner regarding UVic's outage tomorrow and the general bugs we've been facing lately until we get the endpoints figured out.

## Screenshots
image
image

## General Comments

Should this be phrased differently? Not really sure if CourseUp has had messages like these in the past?


liamshatzel (approved):
LGTM

PR #435, opened 2023-01-11
Specify seat error

szeckirjr:
# Description

When the /seats endpoint fails, it removes all other info about a course even if it exists. This PR moves the error message into the component so that the rest still loads for the user.

### Before
image

### After
image

## Checklist

- [ ] The code is documented.


JoshoTheMosho (approved):
These changes look good to me

PR #446, opened 2023-03-01
Update constants.ts

szeckirjr:
# Description

Add summer term

### DIFF
--- a/functions/src/constants.ts
+++ b/functions/src/constants.ts
@@ -9,7 +9,8 @@ export type Term =
   | '202201'
   | '202205'
   | '202209'
-  | '202301';
+  | '202301'
+  | '202305';
 
 // add some code?
 const buildings: { [key: string]: string } = {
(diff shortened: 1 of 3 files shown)


JoshoTheMosho:
On the deployment, it doesn't look like the courses are loaded? https://courseup-cvg5izgab-vikelabs.vercel.app/scheduler/202305

Will also need to update: https://github.com/VikeLabs/courseup/blob/szeckirjr-patch-1/src/common/header/components/TermSelect.tsx

aomi:
> On the deployment, it doesn't look like the courses are loaded? https://courseup-cvg5izgab-vikelabs.vercel.app/scheduler/202305
> 
> Will also need to update: https://github.com/VikeLabs/courseup/blob/szeckirjr-patch-1/src/common/header/components/TermSelect.tsx

backend doesn't have preview builds

aomi (approved):

PR #461, opened 2023-04-01
Fix timezone bug

### LINKED ISSUE #399: Bug: Timetable uses incorrect time when in different time zone
**Describe the Bug/Issue**
The timetable is currently displaying the wrong time for each class for me. For example, CSC 375 starts at 8:30 am but on my timetable it's showing it starts at 9:30 (see screenshot). I assume this is because I'm in Calgary and so the hour is always one ahead.

Not too much of an issue since all courses are offset by the same amount, but for more extreme time zone differences this may look strange.

**To Reproduce**

Steps to reproduce the behaviour/bug/issue:

1. Register for CSC 375 for Fall 2022 in a different time zone (might be able to simulate this)

**Expected Behaviour**

CSC 375 should be displayed as starting at 8:30, not 9:30

**Screenshots**

courseup_bug

**Desktop/Mobile**

 - OS: Windows 8
 - Browser: Brave

keithradford:
# Description

Fixes the timezone bug by forcing everyone into pacific time.

Closes #399

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -16,6 +16,7 @@
         "@sentry/tracing": "^6.16.1",
         "algoliasearch": "^4.11.0",
         "date-fns": "^2.27.0",
+        "date-fns-tz": "^2.0.0",
         "firebase": "^8.9.1",
         "framer-motion": "^6.2.8",
         "history": "^5.0.0",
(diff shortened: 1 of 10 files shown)


szeckirjr (approved):
LGTM! :shipit: this makes a ton more sense for people planning their courses from abroad

PR #467, opened 2023-04-26
display seat info on timetable sidebar

malcolmseyd:
# Description

This change displays seat info on the timetable page. I found this to be a usability improvement when creating a schedule after the registration date, as most sections had filled up but I had to tab back and forth to the Courses tab to see which ones were full.

## Screenshots

The timetable view on a full-sized 1080p monitor (new vs old):
### New
image
### Old
image

The timetable view on a scaled 1080p laptop display (new vs old):
### New
image
### Old
image

## Checklist

- [ ] The code is documented.

## General Comments

I'm not too happy with how the change squishes the rest of the columns on smaller devices. I'm open to suggestions on how to improve that, my current strategy was just to make every existing column smaller.

### DIFF
--- a/src/pages/scheduler/components/SchedulerSections.tsx
+++ b/src/pages/scheduler/components/SchedulerSections.tsx
@@ -1,6 +1,6 @@
 import React, { useCallback, useEffect, useMemo } from 'react';
 
-import { Radio, RadioGroup, Box, HStack, Text, VStack, Tooltip, forwardRef } from '@chakra-ui/react';
+import { Radio, RadioGroup, Box, HStack, Text, VStack, Tooltip, forwardRef, Badge, Flex } from '@chakra-ui/react';
 
 import { MeetingTimes, Section } from 'lib/fetchers';
 import { useDarkMode } from 'lib/hooks/useDarkMode';
@@ -102,12 +102,15 @@ export function SectionGroup({ sections, type, course, courses, handleChange }: 
 
   return (
     <RadioGroup onChange={onChange} value={section} name={type}>
-      {filteredSections.map(({ sectionCode, meetingTimes, additionalNotes }) => (
+      {filteredSections.map(({ sectionCode, meetingTimes, additionalNotes, seats }) => (
         <Option
-          sectionCode={sectionCode}
-          meetingTimes={meetingTimes}
-          additionalNotes={additionalNotes}
-          key={sectionCode}
+          {...{
+            sectionCode,
+            meetingTimes,
+            additionalNotes,
+            seats: seats,
+            key: sectionCode,
+          }}
         />
       ))}
     </RadioGroup>
@@ -129,18 +132,23 @@ export interface OptionsProps {
    * Additional info like section restrictions, etc
    */
   additionalNotes?: string;
+
+  seats: Section['seats'];
 }
 
 const maxAdditionalNotesLength = 200;
 
 export const Option = forwardRef<OptionsProps, 'div'>(
-  ({ meetingTimes, sectionCode, additionalNotes }: OptionsProps, ref): JSX.Element => {
+  ({ meetingTimes, sectionCode, additionalNotes, seats }: OptionsProps, ref): JSX.Element => {
     const mode = useDarkMode();
 
     const truncAdditionalNotes =
       (additionalNotes?.length ?? 0) > maxAdditionalNotesLength
         ? additionalNotes?.substring(0, maxAdditionalNotesLength).trim() + '…'
         : additionalNotes;
+
+    const sectionFull = seats?.enrollment === seats?.maxEnrollment;
+    const waitlistFull = seats?.waitCount === seats?.waitCapacity;
 
     return (
       <Tooltip label={truncAdditionalNotes} isDisabled={!additionalNotes} placement="left">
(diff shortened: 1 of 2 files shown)


src/pages/scheduler/components/SchedulerSections.tsx
```diff
@@ -164,16 +172,30 @@ export const Option = forwardRef<OptionsProps, 'div'>(
...
                     <Text key={time}>{time}</Text>
                   ))}
                 </Box>
-                <Box w="20%" minW="13%">
+                <Box w="10%" minW="7.5%">
                   {m.days}
                 </Box>
-                <Box w="47%">
+                <Box w="40%">
                   <Location short={`${m.buildingAbbreviation} ${m.roomNumber}`} long={m.where} />
```
keithradford, line 199: If you add the `alwaysShort` prop as `true` to this it could help clear up space and make things less squished?

malcolmseyd: done, i also adjusted the spacing of some columns to make things less squishy

malcolmseyd:
I made the location always short, here's a new preview on a laptop display
image

malcolmseyd:
I really like the modified section card. The idea for this change is that you can tell at a glance which sections you should consider, and I feel like hovering makes it hard to compare lots of sections at once.

If I do that change, would you prefer that I implement that exact design? And if so, could I have the source for those icons?

szeckirjr:
> If I do that change, would you prefer that I implement that exact design? And if so, could I have the source for those icons?

Not necessarily! Feel free to explore with whatever design you think makes most sense, that was just a quick idea.

In terms of the icons, those are just placeholders - you should be able to use the [Chakra UI Icons](https://chakra-ui.com/docs/components/icon#all-icons) like Calendar, Clock, but if you need more (like person, which doesn't seem like exist in chakra) you should be able to use _any_ of the ones from [React Icons library](https://react-icons.github.io/react-icons/)

malcolmseyd:
I took a shot at implementing your design. Here's what it looks like right now:

Laptop L (1440px):
image

Laptop (1024px, anything smaller is "mobile" and hides the sidebar"):
image

Long building names wrapped so I chose to use abbreviations (i.e. ECS was "Engineering and Computer Science Building"). In practice, users can use the tooltip if they really don't know what the building is called.

What do you think?

keithradford:
idk why but the icons gave me flashbacks to this
Screen Shot 2023-05-17 at 4 56 50 PM

szeckirjr:
Ohh I like it !!! 🎉 I do like this card-like idea a lot.

I played around with it a bit and I had a couple points:
- add a divider between each section just so we know when the info of one ends and one begins? the full white background makes it hard to discern between sections
- remove the parentheses in the waitlist count (I think it was being used before to represent just that but I don't know if we need it now that a label is shown)
- right now for classes with irregular labs, it generates a new card for _each_ occurrence... maybe there's a way of mapping through the instances and only mapping the calendar + time for each?
image

Good work though, this looks great! :) All these are ideas too btw, so open to suggestions if you disagree with it

malcolmseyd:
Hey, I implemented your recommended changes. Here's what it looks like now:

Laptop L (1440px)
image

Laptop (1024px)
image

I figured using flexbox would make better use of bigger screens and I'm pretty happy with how the Laptop L screenshot looks. Laptop still looks a bit cramped but overall the section options on small screens look way better than they did before.

Future work could include making the meeting times look separated but honestly, I'm pretty happy with it as-is. Please let me know if you have any more feedback.

szeckirjr (approved):
Those cards look amazing! The one thing I noticed is that the hover text is showing the extra text instead of parsing it after "Section information text"

I am not sure why since I assumed it should be getting labels from the same place as prod, so you can try looking into it if you changed where you're getting the text from, otherwise we can see what it looks like in staging 🤷‍♂️

This branch
image

Expected
image

Other than that this looks good to go for me!! Good work thanks for doing this :D

szeckirjr (approved):
Those cards look amazing! The one thing I noticed is that the hover text is showing the extra text instead of parsing it after "Section information text"

I am not sure why since I assumed it should be getting labels from the same place as prod, so you can try looking into it if you changed where you're getting the text from, otherwise we can see what it looks like in staging 🤷‍♂️

This branch
image

Expected
image

Other than that this looks good to go for me!! Good work thanks for doing this :D

malcolmseyd:
I enabled the alwaysShort prop on the Location component a la [this suggestion](https://github.com/VikeLabs/courseup/pull/467#discussion_r1178475650), so thats why the building label changed. I hope that clears it up.

src/pages/scheduler/components/SchedulerSections.tsx
```diff
@@ -129,58 +134,100 @@ export interface OptionsProps {
...
+export const Option = forwardRef<OptionsProps, 'div'>(function Option(
+  { meetingTimes, sectionCode, additionalNotes, seats }: OptionsProps,
+  ref
+): JSX.Element {
+  const mode = useDarkMode();
+
+  const truncAdditionalNotes =
+    (additionalNotes?.length ?? 0) > maxAdditionalNotesLength
+      ? additionalNotes?.substring(0, maxAdditionalNotesLength).trim() + '…'
+      : additionalNotes;
```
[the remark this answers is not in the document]
malcolmseyd: Done. I've implemented something similar in my latest commit.

malcolmseyd:
@szeckirjr I fixed the tooltip issue, it should be truncated appropriately now.

PR #468, opened 2023-04-27
Fix re-render loop caused by `useTerm`

logankuzyk:
# Description

There was a re-render loop happening when the term was included in the URL parameter. Aomi pointed out that the issue was caused by the fact that the hook would use its own `setTerm` function to update the term when the URL parameter was there. I updated the logic so it now would use the URL parameter as the 'first priority' default value instead of setting it after the fact.

## Checklist

- [x] The code is documented.

### DIFF
--- a/src/lib/hooks/useTerm.ts
+++ b/src/lib/hooks/useTerm.ts
@@ -7,11 +7,9 @@ import useLocalStorage from './storage/useLocalStorage';
 
 export const useTerm = (): [Term, (term: string) => void] => {
   const { term: termParam } = useParams();
-  const defaultTerm = getCurrentTerm();
-  const [term, setTerm] = useLocalStorage<string>('user:term', defaultTerm) as [Term, (term: string) => void];
-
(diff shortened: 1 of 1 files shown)


src/lib/hooks/useTerm.ts
```diff
@@ -7,11 +7,9 @@ import useLocalStorage from './storage/useLocalStorage';
 
 export const useTerm = (): [Term, (term: string) => void] => {
   const { term: termParam } = useParams();
-  const defaultTerm = getCurrentTerm();
-  const [term, setTerm] = useLocalStorage<string>('user:term', defaultTerm) as [Term, (term: string) => void];
-
   // The URL parameter should override getCurrentTerm() and the stored term.
-  if (termParam) setTerm(termParam);
+  const defaultTerm = termParam || getCurrentTerm();
```
aomi, line 11: not a huge deal but wrapping this in a `useMemo` would be clutch.

szeckirjr (approved):
:shipit:

PR #488, opened 2023-07-08
feat: Auto update algolia index

### LINKED ISSUE #444: feature: Populate search index (Algolia) via a GitHub Action
Currently, the search feature uses Algolia under the hood. 

In order to capture new courses, the search index must be updated periodically otherwise new courses will not show up in search while showing up when manually navigating courses.

Each record in the index should be shaped like the following:
```json
{
      "subjectCode": "PAAS342",
      "subject": "PAAS",
      "code": "342",
      "pid": "BJm2a_pXV",
      "description": "For students who wish to expand their working vocabulary and develop their skills in reading modern Japanese. Course content may vary from year to year.",
      "title": "Advanced Readings in Japanese II"
}
```
https://www.algolia.com/doc/api-client/methods/indexing/

Records can either be added or updated. Each record in the index requires a unique id. If an id is given when creating records in the index it will be used, otherwise it will be given a generated one.

https://www.algolia.com/doc/api-reference/api-methods/save-objects/?client=javascript

The script should take the current courses (hitting the courseup API probably) and transform the values to match the above object shape. Whether to update records or purge the index and create records from scratch will require some work and research on the developer.

Historically, the `pid` is usually consistent between terms but it's not a guarantee. As such it may make sense to add `term` to list of attributes indexed, this would allow searching between terms and do things like searching for courses via extended attributes like `class_in_session` etc.

szeckirjr:
# Description

This PR adds a TS script to generate a new algolia index and update it so that search has latest data. I got it to work with my own algolia account and it seems to be working if those are the only fields needed:

Closes #444 

image

## Checklist

- [ ] The code is documented.

### DIFF
--- a/.github/workflows/update-algolia.yaml
+++ b/.github/workflows/update-algolia.yaml
@@ -1,0 +1,68 @@
+name: Script - upserts course information to Algolia
+on:
+  worflow_dispatch:
+    inputs:
+      environment:
+        description: Application environment to execute in.
+        required: true
+      term:
+        description: Academic term to execute descript with. ie 202009, 202001
+        required: true
+jobs:
+  lint:
+    name: Lint
+    runs-on: ubuntu-latest
+    defaults:
+      run:
+        working-directory: ./functions
+    steps:
+      - uses: actions/checkout@v2
+      - name: Setup Node.js
+        uses: actions/setup-node@v1
+        with:
+          node-version: 14
+      - name: Cache npm dependencies
+        uses: actions/cache@v2
+        with:
+          key: npm-${{ hashFiles('package-lock.json') }}
+          path: ~/.npm
+          restore-keys: |
+            npm-
+      - name: Install dependencies
+        run: npm ci --ignore-scripts --no-audit --no-progress
+      - name: Lint
+        run: npm run lint
+  update-algolia:
+    name: Update Algolia index
+    needs: [lint]
+    runs-on: ubuntu-latest
+    environment: ${{ github.event.inputs.environment }}
+    defaults:
+      run:
+        working-directory: ./functions
+    steps:
+      - uses: actions/checkout@v2
+      - name: Setup Node.js
+        uses: actions/setup-node@v1
+        with:
+          node-version: 14
+      - name: Cache npm dependencies
+        uses: actions/cache@v2
+        with:
+          key: npm-${{ hashFiles('package-lock.json') }}
+          path: ~/.npm
+          restore-keys: |
+            npm-
+      - name: Install dependencies
+        run: npm ci --ignore-scripts --no-audit --no-progress
+      - name: Set up Cloud SDK
+        uses: google-github-actions/setup-gcloud@v0
+        with:
+          service_account_key: ${{ secrets.FIREBASE_SERVICE_ACCOUNT }}
+          export_default_credentials: true
+      - name: Generate latest course list
+        run: node ./scripts/generate-courses.js ${{ github.event.inputs.term }}
+      - name: Generate index for Algolia
+        run: node ./scripts/generate-algolia-index.js
+      - name: Upload index to Algolia
+        run: node ./scripts/upload-algolia-index.js
--- a/.github/workflows/update-algolia.yaml
+++ b/.github/workflows/update-algolia.yaml
@@ -1,12 +1,12 @@
 name: Script - upserts course information to Algolia
 on:
-  worflow_dispatch:
+  workflow_dispatch:
     inputs:
       environment:
         description: Application environment to execute in.
         required: true
       term:
-        description: Academic term to execute descript with. ie 202009, 202001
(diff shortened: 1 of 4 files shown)


functions/package.json
```diff
@@ -43,6 +43,7 @@
     "@types/ioredis": "^4.27.1",
     "@types/jest": "^26.0.24",
     "@types/node": "^14.18.3",
+    "@types/node-fetch": "^2.6.4",
```
aomi, line 46: I think you need to have the Algolia package in the function package.json for it to work.

.github/workflows/update-algolia.yaml
```diff
@@ -0,0 +1,64 @@
...
+        uses: actions/cache@v2
+        with:
+          key: npm-${{ hashFiles('package-lock.json') }}
+          path: ~/.npm
+          restore-keys: |
+            npm-
+      - name: Install dependencies
+        run: npm ci --ignore-scripts --no-audit --no-progress
+      - name: Lint
+        run: npm run lint
```
aomi, line 34: I know this is done in the other scripts but probs fine removing it.

functions/scripts/update-algolia-index.ts
```diff
@@ -0,0 +1,78 @@
...
+  }));
+
+  index
+    .saveObjects(coursesForIndex)
+    .then(({ objectIDs }) => {
+      console.log(objectIDs);
+    })
+    .catch((err) => {
+      console.log(err);
+    });
```
aomi, line 75: Are you going to update a single index by clearing it and then adding new data or doing this? iirc if you say run this twice in succession you'll have duplicate records in the index. I think the `objectID` should make it so it doesn't but just want to check in.

aomi: What is the plan for stale courses? ie. courses that don't exist anymore cause this will make it so some will linger.

.github/workflows/update-algolia.yaml
```diff
@@ -0,0 +1,64 @@
...
+          path: ~/.npm
+          restore-keys: |
+            npm-
+      - name: Install dependencies
+        run: npm ci --ignore-scripts --no-audit --no-progress
+      - name: Set up Cloud SDK
+        uses: google-github-actions/setup-gcloud@v0
+        with:
+          service_account_key: ${{ secrets.FIREBASE_SERVICE_ACCOUNT }}
+          export_default_credentials: true
```
aomi, line 62: This can be removed if it's not accessing Firestore directly.

functions/scripts/update-algolia-index.ts
```diff
@@ -0,0 +1,78 @@
...
+
+const term: string = process.argv[2];
+
+if (!/20\d{2}0[1,5,9]/.test(term.trim()))
+  throw Error('Invalid term argument format');
+
+const BASE = 'https://courseup.vikelabs.ca/api';
+
+const client = algoliasearch('ALGOLIA_APP_ID', 'ALGOLIA_ADMIN_KEY');
+const index = client.initIndex('ALGOLIA_INDEX_NAME');
```
aomi, line 14: Do these not need to be `process.env` or something?

functions/scripts/update-algolia-index.ts
```diff
@@ -0,0 +1,78 @@
...
+        vals.push({ ...courseDetails });
+      }
+    });
+
+  await Promise.allSettled(workers);
+
+  const coursesForIndex = vals.map((course) => ({
+    objectID: `${course.subject}${course.code}`,
+    subject: course.subject,
+    code: course.code,
```
aomi, line 62: Having a field that's just the concatenation like `objectID` helps with cases where people enter `CSC100` without a space.

aomi (commented):
Also, you'll want to set up synonyms for stuff like `ELEC == ECE` and `"computer science" == csc` etc.,  (people do search sometimes spelling the entire subject 🤷 )

functions/scripts/update-algolia-index.ts
szeckirjr: is `objectID` not taken into consideration for the search?

szeckirjr: and re: synonyms for courses, are you thinking a sort of look up table with pre defined synonyms for the most common stuff?

functions/scripts/update-algolia-index.ts
szeckirjr: yep you're right! I ran this twice and I don't get duplicate records because of the `objectID` field. What it does is replace _all_ other fields in the object aside from the ID (which in our case will be the subject+code)

but hmmm that's a good point 🤔 wiping it clear and _then_ populating might be good but then it worried me about anything going wrong with the API (e.g. running the script with 202401 currently will return 0 courses to be added since API is not working, so it would simply wipe the index instead of not touching it at all). What do you think?

PR #506, opened 2023-11-01
sort search results

### LINKED ISSUE #505: Feature: Lexicographically sort search results by course code
**Describe the Bug/Issue**
Currently the course search results are not sorted by course code. As a user, I expect the course search results to be sorted lexicographically.

**Screenshots**
Screenshot 2023-10-31 at 19-08-18 CourseUp

**Expected Behaviour**
Screenshot 2023-10-31 at 19-10-28 CourseUp

zakariyavkhan:
# Description

Sort the course search results by subject and course number.

Closes #505 

### Before

Screenshot 2023-10-31 at 19-08-18 CourseUp

### After

Screenshot 2023-10-31 at 19-10-28 CourseUp

## Checklist

- [x] The code is documented.

### DIFF
--- a/src/common/layouts/sidebar/components/SearchResults.tsx
+++ b/src/common/layouts/sidebar/components/SearchResults.tsx
@@ -21,9 +21,17 @@ const SearchResults = ({ hits, term }: Props) => {
   const smallScreen = useSmallScreen();
   const navigate = useNavigate();
 
+  // Sort the search hits by subject, then course code
+  const sortedHits = hits.sort((a, b) => {
+    if (a.subject === b.subject) {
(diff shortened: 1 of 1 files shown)


aomi:
I think to do this feature well, it will need to be considered carefully. Right now the search is all handled by Algolia with some minor tweaks to make it easier to use. 

I can think of some situations where this breaks. Since this is anything search, if there's results that are for different subjects it will mess with them. If you are searching for a course within the same subject it messes with the order.

I would consider only engaging this sorting behaviour when there's explicit conditions met like the user is searching for just the subject. Just my two cents. Search is hard :|

Issue #513, opened 2024-03-01
Bug: Summer 2024 Missing


FlyteWizard:
**Describe the Bug/Issue**

Summer 2024 is Missing.

**Expected Behaviour**

The Summer 2024 schedule to be available.

**Additional Context**

It would be nice to have the Summer 2024 schedule out, even if it doesn't include location or professors, so we can easily manipulate scheduling.

PR #515, opened 2024-06-06
[FIX] Implement BANNER API course database population

MNThomson:
# Description

Courseup currently does not have any data for [202405](https://courseup.vikelabs.ca/calendar/202405/) and beyond due to _some UVic api changes_. Luckily there is a [banner api](https://banner.uvic.ca/StudentRegistrationSsb/ssb/searchResults/searchResults?txt_term=202405&pageMaxSize=10000) that just gets all that data for you!
```
curl -v --cookie "JSESSIONID=blahbahblahblah" \
https://banner.uvic.ca/StudentRegistrationSsb/ssb/searchResults/searchResults?txt_term=202405&pageMaxSize=10000
```
(NOTE: there is some scuffed session stuff going on, so before you can run the above command, you have to go to https://banner.uvic.ca/StudentRegistrationSsb/ssb/classSearch/classSearch in your browser and click `Search` with all of the filter boxes empty)

Now, the above API does require a `JSESSIONID` cookie.....oh no! Obviously no one wants to have their cookie stored somewhere that other people can retrieve it. So the CI workflow called `populate-database` now takes an additional [input parameter](https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#onworkflow_dispatchinputs): `cookie`. After use, the `logoff` endpoint is called to invalid the session cookie. Woot woot, no cookies leaked!

**Next Steps:**
- Wrangle the API response into the expected `Sections[]` datatype that was already in use
- Unhackify the implementation to something that's merge-able

fixes: #513 

## Checklist

- [ ] The code is documented.

## General Comments

I wanted to procrastinate an assignment yesterday so I just did this.

Now, I have absolutely no clue if just populating the db is enough to get CU working for future semesters. I don't know if the course scraper is being used in different contexts and not just hitting the firebase db. This is where someone else will have to provide guidance.

Is this a PR that the @VikeLabs/courseup team will support? Should I continue development on this?

### DIFF
--- a/.github/workflows/populate-courses.yaml
+++ b/.github/workflows/populate-courses.yaml
@@ -8,6 +8,9 @@ on:
       term:
         description: Academic term to execute descript with. ie 202009, 202001
         required: true
+      cookie:
+        description: JSESSIONID Cookie for the '/StudentRegistrationSsb' Path
+        required: true
 jobs:
   lint:
     name: Lint
@@ -41,6 +44,7 @@ jobs:
       run:
         working-directory: ./functions
     steps:
+      - run: echo "::add-mask::${{ github.event.inputs.cookie }}"
       - uses: actions/checkout@v2
       - name: Setup Node.js
         uses: actions/setup-node@v1
@@ -61,4 +65,4 @@ jobs:
           service_account_key: ${{ secrets.FIREBASE_SERVICE_ACCOUNT }}
           export_default_credentials: true
       - name: Execute Script
-        run: npm run db:populate ${{ github.event.inputs.term }}
+        run: npm run db:populate "${{ github.event.inputs.term }}" "${{ github.event.inputs.cookie }}"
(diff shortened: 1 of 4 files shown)


szeckirjr:
this is super sweet!! 👀 heads up there's a banner client with some types already under [`functions/src/banner`](https://github.com/VikeLabs/courseup/tree/main/functions/src/banner) too that might be worth checking out, I think it does some of the more basic stuff for querying banner data.

It's been such a transition time with no people knowledgeable with the processes (especially related to backend/data flow/scraper stuff), so we didn't really get anywhere trying to solve the errors we were getting when populating the DB for past couple of terms. I _believe_ having the data in the DB is enough, but I don't understand a lot of that end of the process to say for sure tbh. IIRC there was either missing data from banner or some other error that didn't let us run the populate script for previous courses

MNThomson:
> dont think it will be enough. the database isnt used that much.
> when it pulls only courses in session it will hit the db otherwise it hits the scraper for that term

from @aomi on Discord

Sounds like there would be significant work to the backend and I (with my limited knowledge) see two paths forward:
1. Every request hits the database. This would probably be  expensive but is also probably the "proper answer"
2. The JSON blob with course data is hardcoded into courseup and then periodically updated by volunteers using their own cookie

Not sure this is something I would be able to spearhead with my limited technical knowledge of how the backend of CU works. I'm happy to help out in the future if someone else wants to take a lead on this.

